PR #2, opened 2017-10-06
Initial commit from private repo

CMajeri:
Base building blocks for keycloak bridge daemon.
Some features are not necessary for keycloak but can be used as templates for future daemons.


braoru:
Welcome in the real world :)

PR #13, opened 2019-03-22
New endpoints to grab the events

Fratt:
Added endpoints I'll need to grab the Events from the backoffice.
Feel free to change the signatures, but warn me :-)


harture:
The swagger API has been updated.
The Swagger update according the implementation will be part of PR #31

PR #17, opened 2019-04-04
New Bridge endpoints to grab customers or team members

fperot74:


pkg/management/component.go
```diff
@@ -267,22 +266,22 @@ func (c *component) GetUsers(ctx context.Context, realmName string, paramKV ...s
 
 		if userKc.Attributes != nil {
 			var m = *userKc.Attributes
-
```
fperot74: Could someone tell me if these new lines are added in your IDE too with gofmt ?

PR #18, opened 2019-04-05
Authorisation

harture:
Implementation of authorization enforcement


api/management/api.go
```diff
@@ -12,6 +12,7 @@ type UserRepresentation struct {
 	Label         *string `json:"label,omitempty"`
 	Gender        *string `json:"gender,omitempty"`
 	BirthDate     *string `json:"birthDate,omitempty"`
+	Groups         *[]string `json:"group,omitempty"`
```
fperot74: This line seems not to be aligned like others... Formatting issue ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -151,6 +152,24 @@ func main() {
 		errc <- fmt.Errorf("%s", <-c)
 	}()
 
+	// Authorizations
+	var authorizations management.Authorizations
+	{
+		json, err := ioutil.ReadFile(authorizationConfigFile)
+
+		if err != nil {
+			logger.Log("msg", "could not read YAML authorisation file", "error", err)
```
fperot74: Typo: authorization ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -151,6 +152,24 @@ func main() {
...
+			return
+		}
+
+		authorizations, err = management.LoadAuthorizations(string(json))
+
+		if err != nil {
+			logger.Log("msg", "could not load authorisations", "error", err)
```
fperot74: Typo: authorization ?

pkg/management/authorisation.go
```diff
@@ -0,0 +1,342 @@
...
+
+		if wildcard || realmAllowed || (targetRealm != "master" && nonMasterRealmAllowed) {
+			return nil
+		}
+	}
+
+	return ForbiddenError{}
+}
+
+// ForbiddenError when an operation is not permitted.
```
fperot74: If we use something like this below sample code, we can avoid adding a case in managementErrorHandler() for each new error and avoid unit tests.
```
var ForbiddenError = HTTPError{
	Status:  http.StatusForbidden,
	Message: "ForbiddenError: Operation not permitted.",
}
```
or less 'pretty' but more generic, something like the CreateForbiddenError(...) like the existing CreateMissingParameterError(...)

api/management/api.go
harture: Done

cmd/keycloakb/keycloak_bridge.go
harture: Done

cmd/keycloakb/keycloak_bridge.go
harture: Done

PR #20, opened 2019-04-08
Events db

bsoniam:
Store audit events in the DB (Maria DB)


pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
-	additionalInfo := "" // all the rest in a JSON?
-	/*if m["type"] == "LOGIN_ERROR" {
-		fmt.Println("The rest of event:")
-		fmt.Println(m)
-	}*/
-
-	_, err := cm.db.Exec(insertEvent, origin, realmName, agentUserID, agentUsername, userID, username, ctEventType, kcEventType, kcOperationType, clientID, additionalInfo)
+	// if ctEventType is not "", then record the events in MariaDB
+	if m["ct_event_type"] != "" {
+		origin := "keycloak" // for the moment only eventss of Keycloak
```
harture: typo: eventss

harture: for all the otheers, the comments is jsut before the line, not at the end. (Just for cosmetic ;) )

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+		realmName := m["realmId"]
```
harture: Would be good to have a comment also for this one

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+		//agentUserID - userId of who is performing an action
+		agentUserID := m["userId"]
+		//agentUsername - username of who is performing an action
+		agentUsername := m["username"] // normally, username can be found in the details key of the map
+		//agentRealmName - realm of who is performing an action
+		agentRealmName := "" // found in authdetails
+		//userID - ID of the user that is impacted by the action
+		userID := "" // found in authdetails, in resourcePath
+		//username - username of the user that is impacted by the action
+		username := ""                    //
```
harture: // seems not needed

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+		ctEventType := m["ct_event_type"] // the ct event type is established before at the component level
```
harture: Move it on top of the line, would be good for cosmetic ;)

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+		kcEventType := m["type"]
```
harture: A comment for each next ones would be good

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
-	clientID := m["clienId"]
-	additionalInfo := "" // all the rest in a JSON?
-	/*if m["type"] == "LOGIN_ERROR" {
-		fmt.Println("The rest of event:")
-		fmt.Println(m)
-	}*/
-
-	_, err := cm.db.Exec(insertEvent, origin, realmName, agentUserID, agentUsername, userID, username, ctEventType, kcEventType, kcOperationType, clientID, additionalInfo)
+	// if ctEventType is not "", then record the events in MariaDB
+	if m["ct_event_type"] != "" {
```
harture: would be easier to read, if you consider the case where == "" , then return nil.

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+		kcOperationType := m["operationType"]
+		clientID := m["clientId"]
 
-	if err != nil {
-		//TODO: how is this error treated further?
-		return err
+		//userId is in the resourcePath
+		if resourcePath, ok := m["resourcePath"]; ok {
+			reg := regexp.MustCompile(`[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}`)
+			if strings.HasPrefix(resourcePath, "users") {
```
harture: Is it possible to simplify it with? :
if m["resourcePath"] == "users"{
//....
}

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+		}
+
+		//check if the key authdetails is present
+		if authDetails, ok := m["authDetails"]; ok {
+			eventAuthDetails := []byte(authDetails)
+			var h map[string]string
+			err := json.Unmarshal(eventAuthDetails, &h)
+
+			if err != nil {
+				fmt.Println(err)
```
harture: fmt should not be used I think. we should use the logger instead.

pkg/event/module.go
```diff
@@ -154,28 +159,108 @@ func NewEventsDBModule(db DBEvents) EventsDBModule {
...
+				return err
+			}
+
+			// in authdetails part we can retrieve the client id, agent realm id, agent user id
```
harture: What do you think about of the following alternative ? Would it be easier to read/maintain or not 
```suggestion
for k, v := range h {
  switch k {
    case "clientId":
      clientId = h["clientId"]
    case "//////":
      ///////
    default:
      infoMap[k] = v
  }
}
```

pkg/event/component.go
```diff
@@ -118,6 +119,76 @@ func (c *adminComponent) AdminEvent(ctx context.Context, adminEvent *fb.AdminEve
 	}
 }
 
+func addCTtypeToEvent(event map[string]string) map[string]string {
+	// add the CTEventType
+
+	//ACCOUNT_CREATED
+	if event["operationType"] == "CREATE" {
+		// check if the resourcePath starts with prefix users
+		if strings.HasPrefix(event["resourcePath"], "users") {
```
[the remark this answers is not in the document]
bsoniam: resourcePath is of the format "users/UUID..." or "roles/...."

pkg/event/module.go
bsoniam: No, because resourcePath has as preffix "users" , followed by an UUID. It can also be "roles" followed also by an UUID

pkg/event/module.go
bsoniam: You are right; the logger is TBD

PR #23, opened 2019-04-11
Management of the account status

fperot74:


api/management/swagger-api_management.yaml
```diff
@@ -225,6 +225,29 @@ paths:
       responses:
         200:
           description: successful operation
+  /realms/{realm}/users/{userID}/status:
+    get:
+      tags:
+      - Users
+      summary: Get the account status for the user
```
harture: I think we should be more precise about it: enabled && has a second factor

PR #24, opened 2019-04-11
Back office events

bsoniam:
Store in the DB the following events: 
API_ACCOUNT_CREATION
API_ACCOUNT_DELETION
GET_DETAILS
LOCK_ACCOUNT
UNLOCK_ACCOUNT
INIT_PASSWORD
& restructuring the event module and component


pkg/event/component.go
```diff
@@ -122,48 +123,44 @@ func (c *adminComponent) AdminEvent(ctx context.Context, adminEvent *fb.AdminEve
...
 		//ACCOUNT_CREATED
 		// check if the resourcePath starts with prefix users
-		if strings.HasPrefix(event["resourcePath"], "users") {
+		if strings.HasPrefix(f["resource_path"], "users") {
 			event["ct_event_type"] = "ACCOUNT_CREATED"
 			return event
 		}
 	case "ACTION":
 		//ACTIVATION_EMAIL_SENT
 		// check if the resourcePath ends with sufix send-verify-email
```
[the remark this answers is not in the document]
fperot74: Just saw that typo is not part of your changes... ;-)

pkg/event/component.go
```diff
@@ -230,36 +234,52 @@ func adminEventToMap(adminEvent *fb.AdminEvent) map[string]string {
...
 	var eventMap = make(map[string]string)
-	eventMap["uid"] = fmt.Sprint(event.Uid())
+	var addInfo = make(map[string]string)
+	// if an event has the ct_event_type set already, the flag avoids rewriting it
+	var doNotSetCTEventType bool = false
+
+	addInfo["uid"] = fmt.Sprint(event.Uid())
+
+	time := epochMilliToTime(event.Time()).UTC()
+	eventMap["audit_time"] = time.Format("2006-01-02 15:04:05.000") //audit_time
```
fperot74: Could we use a constant to avoid writing this format for each Format() call ?

pkg/management/component.go
```diff
@@ -152,19 +167,60 @@ func (c *component) CreateUser(ctx context.Context, realm string, user api.UserR
...
-	locationURL, err := c.keycloakClient.CreateUser(accessToken, realm, userRep)
+	locationURL, err := c.keycloakClient.CreateUser(accessToken, realmName, userRep)
 
 	if err != nil {
 		return "", err
 	}
 
+	//store the API call into the DB
+	var event = make(map[string]string)
+	event["ct_event_type"] = "API_ACCOUNT_CREATION"
```
fperot74: In component.go, we don't want to many business unrelated code in our functions. We should try to have the less "invasive" code as possible here. Could we create functions like :
```
var event map[string]string = events.CreateEventMap("API_ACCOUNT_CREATION")
events.AddEventValues("realm_name", realmName, "user_id", user.Id, "username", user.Username, "origin", "back-office", "audit-time", time.Now().UTC())
```
and somewhere else :
```
func CreateEventMap(apiName string /* and other generic parameters... */) map[string]string {
}
func AddEventValues(eventNfo map[string]string, values string...) {
    // can automatically ignore key/value when value is nil
}
func AddEventTime(eventNfo map[string]string, fieldName string, time time.Time) {
}
```
To illustrate what is worrying me, the initial DeleteUser function is only 2 lines long to do what it is aimed to but is more than 20 lines with code not related to user deletion.

PR #37, opened 2019-04-24
Fix groupid

harture:


pkg/management/component_test.go
```diff
@@ -644,6 +644,7 @@ func TestGetUsers(t *testing.T) {
 
 	var accessToken = "TOKEN=="
 	var realmName = "master"
+	var targetRealmName = "master"
```
fperot74: Would have been better to have 2 distinct names for realmName and targetRealmName to be sure that correct values are used at the correct places

harture: good point

PR #39, opened 2019-04-26
added phoneNumberVerified attribute + refactoring mobilephone to phon…

bsoniam:


pkg/management/component.go
```diff
@@ -216,8 +217,8 @@ func (c *component) CreateUser(ctx context.Context, realmName string, user api.U
 
 	var attributes = make(map[string][]string)
 
-	if user.MobilePhone != nil {
-		attributes["mobilephone"] = []string{*user.MobilePhone}
+	if user.PhoneNumber != nil {
```
fperot74: You often need to do the same conversation from api.UserRepresentation to kc.UserRepresentation
I suggest to do it only once in api/management/api.go like for function ConvertCredential...
(Once it is refactorized, don't worry, coverage should decrease)

harture: Refactoring would be good indeed.

PR #46, opened 2019-05-01
Role group mgmt

harture:


configs/authorization.json
```diff
@@ -435,6 +513,16 @@
           "end_user": {}
         }
       },
+      "GetRolesOfUser": {
```
fperot74: "GetRealmRolesForUser" not removed but not present in this file... Was it missing by error ?

pkg/management/endpoint.go
```diff
@@ -21,6 +21,8 @@ type Endpoints struct {
 	UpdateUser                     endpoint.Endpoint
 	GetUsers                       endpoint.Endpoint
 	CreateUser                     endpoint.Endpoint
+	GetRolesOfUser                 endpoint.Endpoint
+	GetGroupsOfUser                endpoint.Endpoint
 	GetUserAccountStatus           endpoint.Endpoint
 	GetClientRoleForUser           endpoint.Endpoint
 	AddClientRoleToUser            endpoint.Endpoint
```
fperot74: GetRealmRoleForUser endpoint not removed ?

configs/authorization.json
harture: Should be removed, where I missed it ?

pkg/management/endpoint.go
harture: Thanks, fixed

PR #47, opened 2019-05-02
Add Multi language (locale) attribute to an user

bsoniam:
To be tested (RPO): create user call


pkg/management/mock/instrumenting.go
```diff
@@ -35,16 +35,19 @@ func (m *Histogram) EXPECT() *HistogramMockRecorder {
 
 // Observe mocks base method
 func (m *Histogram) Observe(arg0 float64) {
+	m.ctrl.T.Helper()
```
fperot74: You shall use the following command as it seems that linux developpers are not supporting m.ctrl.T.Helper()
`find . -path "*/vendor" -prune -o -path "*/mock/*" -name "*.go" -type f -exec sed -i '/Helper()/d' {} \;`

bsoniam: Done

bsoniam:
> Just missing a small adaption in api_test.go for test coverage

Done

PR #49, opened 2019-05-06
Bugfix convert url

harture:


pkg/management/endpoint.go
```diff
@@ -396,6 +397,7 @@ type LocationHeader struct {
 
 // We are currently using a mapping 1:1 for REST API of Bridge and Keycloak, thus we take a shortcut to convert the location of the resource
 func convertLocationUrl(originalURL string, scheme string, host string) string {
```
fperot74: We don't need to specify a TCP port ?

harture: In theory, we should never have a scenario where it is needed

PR #54, opened 2019-05-08
Fix issue with attributes

harture:
Fix for CLOUDTRUST-1061


pkg/management/component.go
```diff
@@ -272,6 +272,23 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 
 	userRep = api.ConvertToKCUser(user)
 
+	// Merge attributes coming from old user and updated one to not loose anything
```
bsoniam: Merge the attributes coming from the old user representation and the updated user representation in order not to lose anything

bsoniam: This is a suggestion as your comment is a bit fuzzy

bsoniam: Maybe you can also add tests with having empty attributes in the updated user representation ?

harture: Thanks for the improved comments.
Good idea for the unittests, I did it.

PR #58, opened 2019-05-13
Http basic auth

harture:
Add HTTP Basic authentication for event/receiver endpoint


pkg/middleware/authentication.go
```diff
@@ -11,6 +12,73 @@ import (
...
+
+			var matched, _ = regexp.MatchString(`^[Bb]asic *`, authorizationHeader)
+
+			if !matched {
+				logger.Log("Authorization Error", "Missing basic token")
+				httpErrorHandler(context.TODO(), http.StatusForbidden, fmt.Errorf("Missing basic token"), w)
+				return
+			}
+
+			var splitToken = strings.Split(authorizationHeader, "Basic ")
```
fperot74: Or extract with regex **^[Bb]asic (.+)$** ? (avoid branches)

pkg/middleware/authentication.go
```diff
@@ -11,6 +12,73 @@ import (
...
+			var password = tokenSubparts[1]
+
+			// Check password match
+			if password != passwordToMatch {
+				logger.Log("Authorization Error", "Invalid password value")
+				httpErrorHandler(context.TODO(), http.StatusForbidden, fmt.Errorf("Invalid token"), w)
+				return
+			}
+
+			var ctx = context.WithValue(req.Context(), "username", username)
```
fperot74: Won't we have collision risk with username extracted from JWT and stored in context ?

harture: The OIDC middleware and this one will never be executed together. One requiere a Basic whereas the other one a Bearer, so I think there is no risk.
In our specific context, we use OIDC for API call coming from the public API, and Basic authentication for internal calls by Keycloak to push events.

pkg/middleware/authentication.go
harture: I have merged and then checked your comments ... :(
I agree it could be improved, this is also the case in OIDC MW.

PR #62, opened 2019-05-14
Db migration

bsoniam:
Check that the minimum required database schema version is deployed when starting the component (bridge and alerting components). If this is not the case, then the component is stopped with an appropriate error log.
The configuration must also contain a flag, whether to check or not this minimum version. The purpose of this flag is to ignore this check if flyway is not yet used for setting up the database (do not even check whether the table flyway_schema_history exists).

### DIFF
--- a/cmd/keycloakb/keycloak_bridge.go
+++ b/cmd/keycloakb/keycloak_bridge.go
@@ -98,8 +98,12 @@ func main() {
 
 		// Enabled units
 		pprofRouteEnabled = c.GetBool("pprof-route-enabled")
+
+		// Database version checks
 		auditMigrationEnabled  = c.GetBool("db-audit-migration")
+		auditMigrationVersion  = c.GetString("db-audit-migration-version")
 		configMigrationEnabled = c.GetBool("db-config-migration")
+		configMigrationVersion = c.GetString("db-config-migration-version")
 
 		// Influx
 		influxWriteInterval = c.GetDuration("influx-write-interval")
@@ -248,7 +252,6 @@ func main() {
 		// DB migration version
 		// checking that the flyway_schema_history has the minimum imposed migration version
 		// auditevents DB
-		auditMigrationVersion := c.GetString("db-audit-migration-version")
 		if auditMigrationEnabled {
 			// DB schema versioning is enabled
 			if auditMigrationVersion != "" {
@@ -276,7 +279,6 @@ func main() {
 		}
 		// DB migration version
 		// cloudtrust_configurations DB
-		configMigrationVersion := c.GetString("db-config-migration-version")
 		if configMigrationEnabled {
 			if configMigrationVersion != "" {
 				var err error
@@ -672,7 +674,7 @@ func config(logger log.Logger) *viper.Viper {
 	//Storage custom configuration in DB
 	v.SetDefault("config-db", true)
 	database.ConfigureDbDefault(v, "db-config", "CT_BRIDGE_DB_CONFIG_USERNAME", "CT_BRIDGE_DB_CONFIG_PASSWORD")
-	
+
 	v.SetDefault("db-config-migration", false)
 	v.SetDefault("db-config-migration-version", "")
 
@@ -760,7 +762,7 @@ func config(logger log.Logger) *viper.Viper {
 	return v
 }
 
-func checkMigrationVersion(logger log.Logger, dbParams dbConfig, conn events.DBEvents, minMigrationVersion string) error {
+func checkMigrationVersion(logger log.Logger, dbParams database.DbConfig, conn database.CloudtrustDB, minMigrationVersion string) error {
 
 	var flywayVersion string
 	row := conn.QueryRow(`SELECT version FROM flyway_schema_history ORDER BY installed_rank DESC LIMIT 1;`)
--- a/cmd/keycloakb/keycloak_bridge.go
+++ b/cmd/keycloakb/keycloak_bridge.go
@@ -7,6 +7,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"os/signal"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -97,6 +98,8 @@ func main() {
 
 		// Enabled units
 		pprofRouteEnabled = c.GetBool("pprof-route-enabled")
+		auditMigrationEnabled  = c.GetBool("db-audit-migration")
+		configMigrationEnabled = c.GetBool("db-config-migration")
 
 		// Influx
 		influxWriteInterval = c.GetDuration("influx-write-interval")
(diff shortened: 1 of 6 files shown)


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -303,16 +340,6 @@ func main() {
 		defer closer.Close()
 	}
 
-	// Audit events DB.
-	type CloudtrustDB interface {
```
fperot74: This move is not necessary : this part will be removed when using common-service

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +863,63 @@ func config(logger log.Logger) *viper.Viper {
 	return v
 }
 
+func checkMigrationVersion(logger log.Logger, dbParams dbConfig, migrationVersionDBConn CloudtrustDB, minMigrationVersion string) error {
```
fperot74: migrationVersionDBConn seems to be used only locally : you shoud remove it as a parameter and declare it locally

We also can keep the connection open and return it at the end of checkMigrationVersion to avoid to reopen it right after this call in the main function

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +863,63 @@ func config(logger log.Logger) *viper.Viper {
...
+		logger.Log("msg", "could not convert string to int", "error", err)
+		return err
+	}
+	minorFlyway, err := strconv.Atoi(minMajorFlyway[1])
+	if err != nil {
+		logger.Log("msg", "could not convert string to int", "error", err)
+		return err
+	}
+
+	// it is required for the last script version of the flyway is to "bigger" than the required version
```
fperot74: rephrase: is to "bigger" => to be "bigger ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +863,63 @@ func config(logger log.Logger) *viper.Viper {
...
+	if (majorFlyway < majorRequired) || (majorFlyway >= majorRequired && minorFlyway < minorRequired) {
```
fperot74: 3 cases:
* majorFlyway < majorRequired
* majorFlyway == majorRequired && minorFlyway < minorRequired
* majorFlyway > majorRequired

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +863,63 @@ func config(logger log.Logger) *viper.Viper {
...
+		logger.Log("msg", "the DB schema is not up-to-date", "DB", dbParams.Database)
+		return fmt.Errorf("DB schema not up-to-date")
+	}
+
+	err = migrationVersionDBConn.Close()
```
fperot74: This statement should be **defer**ed right after connection (not closed in case of error)

(It is more as a best practice than everything else because if there is an error, application will stop and connection will be closed anyway)

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +863,63 @@ func config(logger log.Logger) *viper.Viper {
...
+	if err != nil {
+		logger.Log("msg", "could not close DB connection", "DB", dbParams.Database, "error", err)
+		return err
+	}
+
+	logger.Log("msg", "DB schema version corresponds to the requirements")
```
fperot74: version fits the requirements ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -864,7 +948,7 @@ func getDbConfig(v *viper.Viper, prefix string) *dbConfig {
 func (cfg *dbConfig) openDatabase() (*sql.DB, error) {
 	var err error
 	var dbConn *sql.DB
-	dbConn, err = sql.Open("mysql", fmt.Sprintf("%s:%s@%s(%s)/%s?time_zone='UTC'", cfg.Username, cfg.Password, cfg.Protocol, cfg.HostPort, cfg.Database))
```
fperot74: Seems to be an unwanted change

cmd/keycloakb/keycloak_bridge.go
bsoniam: Indeed,majorFlyway >= majorRequired && minorFlyway < minorRequired
should be 
majorFlyway == majorRequired && minorFlyway < minorRequired

The third case is alright

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -775,13 +775,17 @@ func config(logger log.Logger) *viper.Viper {
 	// Storage events in DB (read/write)
 	v.SetDefault("events-db", false)
 	configureDbDefault(v, "db-audit-rw")
+	v.SetDefault("db-audit-migration", false)
+	v.SetDefault("db-audit-migration-version", "")
```
fperot74: What if xxx-migration-version is empty and xxx-migration is true ? Should the second parameter be automatically considered as false ?

cmd/keycloakb/keycloak_bridge.go
fperot74: Another idea : call this method after db connection is open in main function and simply use the connection you passed as a parameter without need to close it

cmd/keycloakb/keycloak_bridge.go
fperot74: My bad ! :-D

cmd/keycloakb/keycloak_bridge.go
bsoniam: Very good point: this should not happen; but if it is the case and the version number is missing we cannot do the verification in any case. So, I would answer to your second question with yes: if no version number is given then the check should not be done.
Do you agree?

cmd/keycloakb/keycloak_bridge.go
bsoniam: Changed, thanks

cmd/keycloakb/keycloak_bridge.go
bsoniam: Changed

cmd/keycloakb/keycloak_bridge.go
bsoniam: Changed with the first suggestion

cmd/keycloakb/keycloak_bridge.go
bsoniam: Changed

cmd/keycloakb/keycloak_bridge.go
fperot74: migration=true + migration-version empty:
- we consider that the check is disabled
- we consider that it is an error: the application stops
You can choose... ;-)

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +865,58 @@ func config(logger log.Logger) *viper.Viper {
 	return v
 }
 
+func checkMigrationVersion(logger log.Logger, dbParams dbConfig, minMigrationVersion string) error {
+
+	migrationVersionDBConn, err := dbParams.openDatabase()
+	defer migrationVersionDBConn.Close()
```
fperot74: defer after the *if* statement... you may have an invalid migrationVersionDBConn here (perhaps a nil value)

bsoniam: Thanks

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -836,6 +865,59 @@ func config(logger log.Logger) *viper.Viper {
...
+		return err
+	}
+
+	// compare the two versions of the type x.y (major.minor)
+	minMajorRequired := strings.Split(minMigrationVersion, ".")
+	minMajorFlyway := strings.Split(flywayVersion, ".")
+
+	majorRequired, err := strconv.Atoi(minMajorRequired[0])
+	if err != nil {
+		logger.Log("msg", "could not convert string to int", "error", err)
```
fperot74: Could we 4 different messages for "could not convert string to int"... for example "could not convert majorRequired from string to int" ?

configs/keycloak_bridge.yml
```diff
@@ -39,11 +39,14 @@ keycloak-timeout: 5s
 db-audit-rw-host-port: 127.0.0.1:3306
 db-audit-rw-username: root
 db-audit-rw-password: admin
-db-audit-rw-database: auditevents
+db-audit-rw-database: auditevents2
```
fperot74: May be for your personal use... I think this should not be commited

PR #65, opened 2019-05-16
Remove leakage of sensitive information

harture:
There is some space for improvements and we will have to fix this issue in all our modules

### DIFF
--- a/cmd/keycloakb/keycloak_bridge.go
+++ b/cmd/keycloakb/keycloak_bridge.go
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 	"sort"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -831,7 +841,11 @@ func config(logger log.Logger) *viper.Viper {
 	sort.Strings(keys)
 
 	for _, k := range keys {
-		logger.Log(k, v.Get(k))
+		if _, censored := censoredParameters[k]; censored {
```
fperot74: I would add another condition : if "password" is found in the parameter name, it is automatically censored (and perhaps some other keywords)

harture: Added

fperot74: With your latest fix, you don't need anymore rows like `censoredParameters["XXX-password"] = true`

fperot74: Another idea: only 2 branches for if statement with 

> censored || strings.Contains(k, "password")

harture:
(Test pour voir si ça marche de répondre dans les mails.)
Je vois le fix comme un fallback mais je penses que on doit continuer a le faire explicitement.

PR #72, opened 2019-05-24
[CLOUDTRUST-1262] Get groups

bsoniam:
Added get groups endpoint 
Added authorization for GetRoles: anyone can query and get the roles of his realm

### DIFF
--- a/cmd/keycloakb/keycloak_bridge.go
+++ b/cmd/keycloakb/keycloak_bridge.go
@@ -598,7 +598,7 @@ func main() {
 		managementSubroute.Path("/realms/{realm}/clients/{clientID}/roles").Methods("POST").Handler(createClientRolesHandler)
 
 		//groups
-		managementSubroute.Path("/realms/{realm:[a-zA-Z0-9_-]+}/groups").Methods("GET").Handler(getGroupsHandler)
+		managementSubroute.Path("/realms/{realm}/groups").Methods("GET").Handler(getGroupsHandler)
 
 		// custom configuration par realm
 		managementSubroute.Path("/realms/{realm}/configuration").Methods("GET").Handler(getRealmCustomConfigurationHandler)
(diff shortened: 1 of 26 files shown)


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -594,6 +597,9 @@ func main() {
 		managementSubroute.Path("/realms/{realm}/clients/{clientID}/roles").Methods("GET").Handler(getClientRolesHandler)
 		managementSubroute.Path("/realms/{realm}/clients/{clientID}/roles").Methods("POST").Handler(createClientRolesHandler)
 
+		//groups
+		managementSubroute.Path("/realms/{realm:[a-zA-Z0-9_-]+}/groups").Methods("GET").Handler(getGroupsHandler)
```
harture: the regexp is not needed anymore here. We manage it in http.go

configs/authorization.json
```diff
@@ -119,6 +129,16 @@
         "master": {
           "integrator_agent": {}
         },
+      "GetGroups": {
+        "master": {
```
harture: I think DEP should also be added

configs/authorization.json
```diff
@@ -119,6 +129,16 @@
...
+          "*": {}
+        }
+      },
+      "GetRoles": {
+        "master": {
```
harture: DEP may be useful

configs/authorization.json
```diff
@@ -245,6 +265,16 @@
           "*": {}
         }
       },
+      "GetGroups": {
+        "master": {
```
harture: should be DEP instead of master

configs/authorization.json
```diff
@@ -245,6 +265,16 @@
...
+          "*": {}
+        }
+      },
+      "GetRoles": {
+        "master": {
```
harture: should be DEP instead of master

configs/authorization.json
bsoniam: DONE

configs/authorization.json
bsoniam: DONE

configs/authorization.json
bsoniam: DONE

cmd/keycloakb/keycloak_bridge.go
bsoniam: DONE

PR #74, opened 2019-05-28
CLOUDTRUST-938 Statistics swagger

Fratt:


Fratt:
I'm really not sure about the case I used (case_i_used).
It seems that sometimes we use this_case, and sometimes thisCase...
I let you decide :-)

fperot74:
Partially done in cloudtrust/keycloak-bridge#80 for some statistics...
Development of other ones will have to be rescheduled later

PR #77, opened 2019-06-04
Use constants for columns in events management

fperot74:
(build will fail until cloudtrust/common-service#15 is merged)

### DIFF
--- a/pkg/account/component.go
+++ b/pkg/account/component.go
@@ -52,7 +52,7 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 	_, err := c.keycloakClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
 
 	//store the API call into the DB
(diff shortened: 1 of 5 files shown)


pkg/account/component.go
```diff
@@ -52,7 +52,7 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 	_, err := c.keycloakClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
 
 	//store the API call into the DB
-	_ = c.reportEvent(ctx, "PASSWORD_RESET", "realm_name", realm, "user_id", userID, "username", username)
+	_ = c.reportEvent(ctx, "PASSWORD_RESET", database.CtEventRealmName, realm, database.CtEventUserID, userID, "username", username)
```
bsoniam: database.CtEventUsername for "username"

fperot74: Fixed 👍

PR #87, opened 2019-06-20
CLOUDTRUST-1337 Rights to use custom configuration page

Fratt:


harture:
Could be good to check if TOE should have this right and update https://confluence.svc.elca.ch/pages/viewpage.action?pageId=366543489

Fratt:
> Could be good to check if TOE should have this right and update https://confluence.svc.elca.ch/pages/viewpage.action?pageId=366543489

This is not in the matrix. As for the TOE, in the user story it mentions "As an application owner or integrator..." The Application Owner is using the TOE account imo.

PR #88, opened 2019-06-21
Init pwd

bsoniam:
added INIT_PASSWORD event when a support user calls the API execute-actions-email with the action sms-password-set

### DIFF
--- a/pkg/management/component.go
+++ b/pkg/management/component.go
@@ -474,8 +474,8 @@ func (c *component) ResetPassword(ctx context.Context, realmName string, userID 
 	}
 
 	//store the API call into the DB
-	// the error should be treated
-	_ = c.reportEvent(ctx, "INIT_PASSWORD", database.CtEventRealmName, realmName, database.CtEventUserID, userID)
+	//_ = c.reportEvent(ctx, "INIT_PASSWORD", database.CtEventRealmName, realmName, database.CtEventUserID, userID)
+	// removed as it is not used
 
 	return nil
 }
@@ -490,9 +490,14 @@ func (c *component) ExecuteActionsEmail(ctx context.Context, realmName string, u
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 
 	var actions = []string{}
+	var init_password_action = "sms-password-set"
 
 	for _, requiredAction := range requiredActions {
 		actions = append(actions, string(requiredAction))
+		if string(requiredAction) == init_password_action {
+			//store the API call into the DB
+			_ = c.reportEvent(ctx, "INIT_PASSWORD", database.CtEventRealmName, realmName, database.CtEventUserID, userID)
+		}
 	}
 
 	return c.keycloakClient.ExecuteActionsEmail(accessToken, realmName, userID, actions, paramKV...)
--- a/pkg/management/component.go
+++ b/pkg/management/component.go
@@ -13,6 +13,10 @@ import (
 	kc "github.com/cloudtrust/keycloak-client"
 )
 
+const (
+	initPasswordAction = "sms-password-set"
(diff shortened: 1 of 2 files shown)


pkg/management/component.go
```diff
@@ -490,9 +490,14 @@ func (c *component) ExecuteActionsEmail(ctx context.Context, realmName string, u
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 
 	var actions = []string{}
+	var init_password_action = "sms-password-set"
```
fperot74: You might have a go-lint recommandation asking not to use underscore in variable names. initPasswordAction would be more Go-compliant

pkg/management/component.go
```diff
@@ -474,8 +474,8 @@ func (c *component) ResetPassword(ctx context.Context, realmName string, userID
 	}
 
 	//store the API call into the DB
-	// the error should be treated
-	_ = c.reportEvent(ctx, "INIT_PASSWORD", database.CtEventRealmName, realmName, database.CtEventUserID, userID)
+	//_ = c.reportEvent(ctx, "INIT_PASSWORD", database.CtEventRealmName, realmName, database.CtEventUserID, userID)
```
fperot74: If not used, you'd better completely remove this block

pkg/management/component_test.go
```diff
@@ -1291,9 +1291,11 @@ func TestExecuteActionsEmail(t *testing.T) {
 
 	// Send email actions
 	{
+
 		mockKeycloakClient.EXPECT().ExecuteActionsEmail(accessToken, realmName, userID, actions, key1, value1, key2, value2).Return(nil).Times(1)
 
 		var ctx = context.WithValue(context.Background(), cs.CtContextAccessToken, accessToken)
+		mockEventDBModule.EXPECT().ReportEvent(ctx, "INIT_PASSWORD", "back-office", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
```
fperot74: We could have 2 calls to ExecuteActionsEmail in this test : it would check that ReportEvent is called only if sms-password-set is sent to ExecuteActionsEmail

pkg/management/component.go
bsoniam: I would prefer to leave it as I don't know if the back office will start using it or not soon

pkg/management/component_test.go
```diff
@@ -1281,8 +1281,8 @@ func TestExecuteActionsEmail(t *testing.T) {
 	var accessToken = "TOKEN=="
 	var realmName = "master"
 	var userID = "1245-7854-8963"
-	var reqActions = []api.RequiredAction{"action1", "action2"}
-	var actions = []string{"action1", "action2"}
+	var reqActions = []api.RequiredAction{"sms-password-set", "action1", "action2"}
```
fperot74: If you move init_password_action declaration in a const (...) block at the beginning of component.go, you could reuse it here and prevent potential mis-spelling.

pkg/management/component.go
fperot74: ok

pkg/management/component.go
bsoniam: Hahaha, only now I noticed how I named this variable (Friday tiredness) :))). 
Thanks, good point.

pkg/management/component_test.go
bsoniam: Done

pkg/management/component.go
harture: If we decide to leave the functionality, I would suggest to also leave the reportEvent.

bsoniam: done

PR #90, opened 2019-06-24
Send reminder email

bsoniam:
Add the new "send onboarding reminder mail" API in the bridge.
At the authorization level: SendReminderEmail is similar to SendVerifyEmail; to be decided if this will stay like this

### DIFF
--- a/pkg/management/endpoint_test.go
+++ b/pkg/management/endpoint_test.go
@@ -599,6 +599,50 @@ func TestSendNewEnrolmentCodeEndpoint(t *testing.T) {
 
 }
 
+func TestSendReminderEmailEndpoint(t *testing.T) {
+	var mockCtrl = gomock.NewController(t)
+	defer mockCtrl.Finish()
+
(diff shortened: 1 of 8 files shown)


pkg/management/endpoint_test.go
```diff
@@ -599,6 +599,50 @@ func TestSendNewEnrolmentCodeEndpoint(t *testing.T) {
...
+		var userID = "123-456-789"
+		var ctx = context.Background()
+		var req = make(map[string]string)
+		var lifespan = 3600
+		req["realm"] = realm
+		req["userID"] = userID
+		req["client_id"] = "123789"
+		req["redirect_uri"] = "http://redirect.com"
+		req["lifespan"] = string(lifespan)
+		req["toto"] = "tutu" // Check this param is not transmitted
```
fperot74: Yes, this parameter should not be transmitted but where do we check so ?

bsoniam: Added a short explanation of this

PR #91, opened 2019-06-24
[CLOUDTRUST-1408] Add Location as a value for CORS Access-Control-Exp…

fperot74:
…ose-Headers


Fratt:
Thanks, it works as expected (on the bridge side)
It's not working though on the Angular side, I'm still figuring out why.
**Don't merge this yet, maybe it won't be needed after all.**

Fratt:
Okay got it, it's working perfectly :-) Thanks, this can be merged!

PR #93, opened 2019-06-26
Invalid password message

bsoniam:
Catch the error message when trying to update the password

### DIFF
--- a/pkg/account/component.go
+++ b/pkg/account/component.go
@@ -51,8 +51,18 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 
 	_, err := c.keycloakClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
 
+	var updateError error
+	switch err.Error() {
+	case "invalidPasswordExistingMessage":
+		updateError = commonhttp.Error{
+			Status:  http.StatusBadRequest,
+			Message: err.Error()}
+	default:
+		updateError = err
+	}
+
 	//store the API call into the DB
 	_ = c.reportEvent(ctx, "PASSWORD_RESET", "realm_name", realm, "user_id", userID, "username", username)
 
-	return err
+	return updateError
 }
--- a/pkg/account/component.go
+++ b/pkg/account/component.go
@@ -51,14 +51,16 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 
 	_, err := c.keycloakClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
 
-	var updateError error
-	switch err.Error() {
-	case "invalidPasswordExistingMessage":
-		updateError = commonhttp.Error{
-			Status:  http.StatusBadRequest,
-			Message: err.Error()}
-	default:
-		updateError = err
+	var updateError error = nil
+	if err != nil {
+		switch err.Error() {
+		case "invalidPasswordExistingMessage":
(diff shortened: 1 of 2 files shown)


pkg/account/component.go
```diff
@@ -51,8 +51,20 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 
 	_, err := c.keycloakClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
 
+	var updateError error = nil
+	if err != nil {
+		switch err.Error() {
+		case "invalidPasswordExistingMessage":
+			updateError = commonhttp.Error{
```
fperot74: You can write it this way : updateError = commonhttp.CreateBadRequestError(err.Error())
But I'm not sure we want to send err.Error() in the response (could leak information as we are not sure what keycloak is returning)

pkg/account/component_test.go
```diff
@@ -51,3 +52,39 @@ func TestUpdatePassword(t *testing.T) {
...
+	ctx = context.WithValue(ctx, cs.CtContextRealm, realm)
+	ctx = context.WithValue(ctx, cs.CtContextUserID, userID)
+	ctx = context.WithValue(ctx, cs.CtContextUsername, username)
+
+	mockKeycloakClient.EXPECT().UpdatePassword(accessToken, realm, oldPasswd, newPasswd, newPasswd).Return("", fmt.Errorf("invalidPasswordExistingMessage")).Times(1)
+	mockEventDBModule.EXPECT().ReportEvent(gomock.Any(), "PASSWORD_RESET", "self-service", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
+
+	err := component.UpdatePassword(ctx, oldPasswd, newPasswd, newPasswd)
+
+	assert.Equal(t, true, err != nil)
```
fperot74: assert.True(t, err!=nil)

pkg/account/component_test.go
```diff
@@ -51,3 +52,39 @@ func TestUpdatePassword(t *testing.T) {
...
+
+	mockKeycloakClient.EXPECT().UpdatePassword(accessToken, realm, oldPasswd, newPasswd, newPasswd).Return("", fmt.Errorf("invalid")).Times(1)
+	mockEventDBModule.EXPECT().ReportEvent(gomock.Any(), "PASSWORD_RESET", "self-service", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
+
+	err = component.UpdatePassword(ctx, oldPasswd, newPasswd, newPasswd)
+
+	assert.Equal(t, true, err != nil)
```
fperot74: assert.True(...)

pkg/account/component.go
bsoniam: Yes, and related to this I did another PR for the Keycloak client, so that we don't send the whole response back but only the errorMessage part (if it is present)

pkg/account/component_test.go
bsoniam: Done, thanks

pkg/account/component_test.go
bsoniam: Done, thanks

pkg/account/component_test.go
fperot74: OMG ! Did I really suggest that ?
assert.NotNil(t, err) would have been better !

PR #101, opened 2019-07-25
CLOUDTRUST-1479 Leveled logs

harture:


configs/keycloak_bridge.yml
```diff
@@ -6,6 +6,13 @@ internal-http-host-port: 0.0.0.0:8888
 management-http-host-port: 0.0.0.0:8877
 account-http-host-port: 0.0.0.0:8866
 
+# Log level
+#   - error: log only error log level
+#   - warn: log error and warn log levels
+#   - info: log error, warn, info log levels
+#   - debug: log error, warn, info, debug log levels
+log-level: debug
```
fperot74: We should not commit the debug value

internal/keycloakb/logging.go
```diff
@@ -0,0 +1,9 @@
+package keycloakb
+
+// Logger interface for logging with level
+type Logger interface {
```
fperot74: Should be defined in common-service ???

pkg/event/logging_test.go
```diff
@@ -30,7 +30,7 @@ func TestMuxComponentLoggingMW(t *testing.T) {
 
 	// Event.
 	mockMuxComponent.EXPECT().Event(ctx, "Event", event).Return(nil).Times(1)
-	mockLogger.EXPECT().Log("unit", "Event", "type", "Event", "correlation_id", corrID, "took", gomock.Any()).Return(nil).Times(1)
+	mockLogger.EXPECT().Info("unit", "Event", "type", "Event", "correlation_id", corrID, "took", gomock.Any()).Return(nil).Times(1)
```
fperot74: Do we really want to mock loggers ? (We could use NoopLogger rather than a mocked logger)

pkg/event/mock_test.go
```diff
@@ -3,6 +3,6 @@ package event
 //go:generate mockgen -destination=./mock/event.go -package=mock -mock_names=MuxComponent=MuxComponent,Component=Component,AdminComponent=AdminComponent,ConsoleModule=ConsoleModule,StatisticModule=StatisticModule github.com/cloudtrust/keycloak-bridge/pkg/event MuxComponent,Component,AdminComponent,ConsoleModule,StatisticModule
 //go:generate mockgen -destination=./mock/dbmodule.go -package=mock -mock_names=EventsDBModule=EventsDBModule github.com/cloudtrust/common-service/database EventsDBModule
 //go:generate mockgen -destination=./mock/instrumenting.go -package=mock -mock_names=Histogram=Histogram,Metrics=Metrics github.com/cloudtrust/common-service/metrics Histogram,Metrics
-//go:generate mockgen -destination=./mock/logging.go -package=mock -mock_names=Logger=Logger github.com/go-kit/kit/log Logger
+//go:generate mockgen -destination=./mock/logging.go -package=mock -mock_names=Logger=Logger github.com/cloudtrust/common-service/log Logger
```
fperot74: Do we really want to mock loggers ? (We could use NoopLogger rather than a mocked logger)

internal/keycloakb/logging.go
harture: In Go, the interface should be defined where you are using it, not in the library so I think it is ok to declare it there.

pkg/event/logging_test.go
harture: As we are testing the logging, sounds good to me to use a mock to ensure we really call the logger, but I agree for the others cases

pkg/event/mock_test.go
harture: Agree

configs/keycloak_bridge.yml
harture: As this config is only used for Dev, why not put debug into ti ?

PR #103, opened 2019-08-08
[CLOUDTRUST-1504] reset password

bsoniam:
adapt the exisiting reset-password endpoint to generate a random password if no value is provided in the request.
The generated value will be returned in the content of the response (JSON structure).
The randomly generated password must comply with the password policy defined in the target realm.

### DIFF
--- a/internal/keycloakb/passwordgeneration.go
+++ b/internal/keycloakb/passwordgeneration.go
@@ -1,7 +1,6 @@
 package keycloakb
 
 import (
-	"fmt"
 	"math/rand"
 	"strconv"
 	"strings"
@@ -48,7 +47,6 @@ func GeneratePasswordFromKeycloakPolicy(policy string, minLength int) (string, e
 				for j := 0; j < minLength; j++ {
 					// make sure that the password has the minimum length required by choosing random lower case letters
 					pwdElems = append(pwdElems, string(lowerCase[rand.Intn(len(lowerCase))]))
-					fmt.Println(pwdElems)
 				}
 			} else {
 				return "", err
@@ -62,38 +60,32 @@ func GeneratePasswordFromKeycloakPolicy(policy string, minLength int) (string, e
 		switch keyValueItem[0] {
 		case "specialChars":
 			// pick randomly special characters from ?!#%$
-			fmt.Println("specialChars")
 			minRequired, err := strconv.Atoi(keyValueItem[1])
 			if err == nil {
 				specialChars := []string{"?", "!", "#", "%", "$"}
 				for j := 0; j < minRequired; j++ {
 					pwdElems = append(pwdElems, specialChars[rand.Intn(len(specialChars))])
-					fmt.Println(pwdElems)
 				}
 			} else {
 				return "", err
 			}
 
 		case "upperCase":
-			fmt.Println("upperCase")
 			minRequired, err := strconv.Atoi(keyValueItem[1])
 			if err == nil {
 				const upperCase = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 				for j := 0; j < minRequired; j++ {
 					pwdElems = append(pwdElems, string(upperCase[rand.Intn(len(upperCase))]))
-					fmt.Println(pwdElems)
 				}
 			} else {
 				return "", err
 			}
 
 		case "digits":
-			fmt.Println("digits")
 			minRequired, err := strconv.Atoi(keyValueItem[1])
 			if err == nil {
 				for j := 0; j < minRequired; j++ {
 					pwdElems = append(pwdElems, strconv.Itoa(rand.Intn(10)))
-					fmt.Println(pwdElems)
 				}
 			} else {
 				return "", err
@@ -103,7 +95,6 @@ func GeneratePasswordFromKeycloakPolicy(policy string, minLength int) (string, e
 	}
 	rand.Shuffle(len(pwdElems), func(i, j int) { pwdElems[i], pwdElems[j] = pwdElems[j], pwdElems[i] })
 	pwd := strings.Join(pwdElems, "")
-	fmt.Println(pwd)
 	return pwd, nil
 
 }
--- a/internal/keycloakb/passwordgeneration.go
+++ b/internal/keycloakb/passwordgeneration.go
@@ -8,7 +8,11 @@ import (
 )
 
 const (
-	lowerCase = "abcdefghijklmnopqrstuvwxyz"
+	lowerCase    = "abcdefghijklmnopqrstuvwxyz"
+	upperCase    = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	specialChars = "?!#%$"
+	digits       = "0123456789"
+	alphabet     = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ?!*0123456789"
 )
 
 // appendCharacters appends a number of characters from a certain alphabet to a string array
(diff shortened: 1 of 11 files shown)


internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
...
+					// make sure that the password has the minimum length required by choosing random lower case letters
+					pwdElems = append(pwdElems, string(lowerCase[rand.Intn(len(lowerCase))]))
+				}
+			} else {
+				return "", err
+			}
+		}
+	}
+
+	rand.Seed(time.Now().Unix())
```
fperot74: Already seeded in main keycloak_bridge.go at line 59

internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
...
+	//pwdReq := string(reg.Find([]byte()))
+
+	f := func(c rune) bool {
+		return !unicode.IsLetter(c) && !unicode.IsNumber(c)
+	}
+
+	for i := 0; i < len(policyItems); i++ {
+		keyValueItem := strings.FieldsFunc(policyItems[i], f)
+		switch keyValueItem[0] {
+		case "length": // the minimum length of the password
```
fperot74: You can merge both `for` statements: we don't need the "length" case to be executed first

internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+			} else {
+				return "", err
+			}
+		}
+	}
+
+	rand.Seed(time.Now().Unix())
+	for i := 0; i < len(policyItems); i++ {
+		keyValueItem := strings.FieldsFunc(policyItems[i], f)
```
fperot74: I like the idea ;-)

internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
...
+	for j := 0; j < minLength; j++ {
+		pwdElems = append(pwdElems, string(alphabet[rand.Intn(len(alphabet))]))
+	}
+
+	pwd := strings.Join(pwdElems, "")
+	return pwd
+}
+
+// GeneratePasswordFromKeycloakPolicy generates a random password respecting the keycloak password policy
+func GeneratePasswordFromKeycloakPolicy(policy string, minLength int) (string, error) {
```
fperot74: minLength parameter is not used. You might consider removing it or using the max value of this minLength and the one converted from the parameter length(...) provided in the given policy

internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// generate a random password that corresponds to the password policy
+	//reg := regexp.MustCompile(`[a-zA-z]+[(]{1}[0-9]+[)]{1}`)
+	//pwdReq := string(reg.Find([]byte()))
+
+	f := func(c rune) bool {
+		return !unicode.IsLetter(c) && !unicode.IsNumber(c)
+	}
+
+	for i := 0; i < len(policyItems); i++ {
```
fperot74: We could first parse the policy to get a map[string]int (only with parameters having a number between parenthesis).
Then we could call a method appendCharacters like this:

> appendCharacters(pwdElems, "abcdefghijklmnopqrstuvwxyz", map, "length")
> appendCharacters(pwdElems, "0123456789", map, "digits")
> appendCharacters(pwdElems, "?!#%$", map, "specialChars")

pkg/management/component.go
```diff
@@ -488,25 +488,55 @@ func (c *component) AddClientRolesToUser(ctx context.Context, realmName, userID,
...
+		//obtain password policy
+		realmKc, err := c.keycloakClient.GetRealm(accessToken, realmName)
+		if err == nil {
+			if realmKc.PasswordPolicy == nil {
+				// no Keycloak password policy impose
+				pwd = internal.GeneratePassword(minLength)
+			} else {
+				policy := *realmKc.PasswordPolicy
+				pwd, err = internal.GeneratePasswordFromKeycloakPolicy(policy, minLength)
+				for pwd == userID {
```
fperot74: Yeaaaaah !!! Good point !

pkg/management/component.go
```diff
@@ -488,25 +488,55 @@ func (c *component) AddClientRolesToUser(ctx context.Context, realmName, userID,
...
 
-	err := c.keycloakClient.ResetPassword(accessToken, realmName, userID, credKc)
+	if password.Value == nil {
+		// no password value was provided; a new password, that respects the password policy of the realm, will be generated
+		var minLength int = 12
+
+		//obtain password policy
+		realmKc, err := c.keycloakClient.GetRealm(accessToken, realmName)
+		if err == nil {
+			if realmKc.PasswordPolicy == nil {
```
fperot74: I would create an internal.GeneratePasswordXXX(policy string*, minLength int) that would try to generate passwords (while pwd==userID AND err==nil) and decide to call GeneratePassword or GeneratePasswordFromKeycloakPolicy according to the given policy. It would be shorter here in component.go

internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
+package keycloakb
+
+import (
```
fperot74: No unit tests here... No necessary mock, they should not be hard to write.

pkg/management/component_test.go
```diff
@@ -1224,11 +1224,74 @@ func TestResetPassword(t *testing.T) {
 			Value: &password,
 		}
 
-		err := managementComponent.ResetPassword(ctx, "master", userID, passwordRep)
+		_, err := managementComponent.ResetPassword(ctx, "master", userID, passwordRep)
 
 		assert.Nil(t, err)
 	}
 
+	// No password offered
```
fperot74: Finally, I have a last suggestion:
In internal/passwordgeneration.go, you can create an interface PasswordGenerator
CreateGenerator(policy *string) check if policy is nil and creates an instance of type defaultPasswordGenerator or of type keycloakPolicyPasswordGenerator. Both types are implementing the interface (with the functions you already wrote)
During the creation of keycloakPolicyPasswordGenerator, you parse the policy to a map[string]int
You can now easiliy generate a password (and re-generate it if password==username) with the best performances.
And here, in the management testing part, you don't have to worry of the generation, just check that the GeneratePassword function of the mock is called

internal/keycloakb/passwordgeneration.go
```diff
@@ -0,0 +1,100 @@
...
+	"strings"
+	"time"
+	"unicode"
+)
+
+// GeneratePassword generates a password of a given length
+func GeneratePassword(minLength int) string {
+
+	var pwdElems []string
+	const alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ?!*0123456789"
```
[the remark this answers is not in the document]
bsoniam: Can you please remind me: which are the ambiguous characters ? 0 vs O, 1 vs I ?

internal/keycloakb/passwordgeneration.go
bsoniam: completely right: at the beginning I could not, but having done some corrections my code simplified

internal/keycloakb/passwordgeneration.go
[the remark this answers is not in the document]
harture: I would say:
* upper-case i, lower-case l, (the number 1), vertical bar |
* upper-case o, the number 0
Seems to be the general agreement on the web

pkg/management/component.go
bsoniam: DONE

internal/keycloakb/passwordgeneration.go
bsoniam: OK

internal/keycloakb/passwordgeneration.go
bsoniam: DONE

pkg/management/component.go
bsoniam: DONE

internal/keycloakb/passwordgeneration.go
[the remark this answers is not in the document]
bsoniam: DONE

internal/keycloakb/passwordgeneration.go
bsoniam: DONE

PR #104, opened 2019-08-14
Self service

harture:


api/account/api.go
```diff
@@ -0,0 +1,92 @@
...
+
+// ConvertCredential creates an API credential from a KC credential
+func ConvertCredential(credKc *kc.CredentialRepresentation) CredentialRepresentation {
+	var cred CredentialRepresentation
+	cred.Id = credKc.Id
+	cred.Type = credKc.Type
+	cred.UserLabel = credKc.UserLabel
+	cred.CreatedDate = credKc.CreatedDate
+	cred.CredentialData = credKc.CredentialData
+	cred.Temporary = credKc.Temporary
```
bsoniam: how about cred.Value?

Gopkg.toml
```diff
@@ -31,7 +31,8 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  version = "1.0-rc3"
+  branch = "credentials"
```
fperot74: You really wanted to merge this file ?

configs/keycloak_bridge.yml
```diff
@@ -68,6 +69,7 @@ db-audit-ro-max-idle-conns: 2
 db-audit-ro-conn-max-lifetime: 10
 db-audit-ro-migration: false
 db-audit-ro-migration-version: 0.1
+db-audit-ro-connection-check: false
```
fperot74: This is just for our test environments, right ? We don't want to use this setting when deployed ?

harture: This is just for our local dev env

Gopkg.toml
harture: Created PR (https://github.com/cloudtrust/keycloak-client/pull/18). Thanks to have notice it.

api/account/api.go
harture: Value will be either used to transmit an info to Keycloak or to store a legacy secret. Thus we never want to return the value to final user.
I will add a comment about it as it not obvious at all.

harture: Value will be either used to transmit an info to Keycloak or to store a legacy secret. Thus we never want to return the value to final user.
I will add a comment about it as it not obvious at all.

PR #119, opened 2019-09-12
[CLOUDTRUST-1649] new password generation method

bsoniam:


pkg/management/component.go
```diff
@@ -534,7 +534,8 @@ func (c *component) ResetPassword(ctx context.Context, realmName string, userID
 	credKc.Type = &passwordType
 
 	if password.Value == nil {
-		// no password value was provided; a new password, that respects the password policy of the realm, will be generated
+		// the commented code respects the following scenario: if the realm has a pwd policy, the generated pwd respects this policy
+		/*// no password value was provided; a new password, that respects the password policy of the realm, will be generated
```
fperot74: /*// is a little bit weird...
It might be more readable with a new line after /*

Gopkg.toml
```diff
@@ -27,11 +27,11 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/common-service"
-  version = "v1.0-rc6"
+  version = "=v1.0-rc6"
```
fperot74: Why = ?
Is it an error or a feature I don't know ?

harture: = in toml means exactly this version. If there is no =, it only means >=.
Here, I think we should remove the =

bsoniam: I added the = as the error handling changes was causing otherwise problems;

pkg/management/component.go
bsoniam: DONE

Gopkg.toml
bsoniam: DONE

PR #121, opened 2019-09-18
Fix update password

bsoniam:


pkg/account/component.go
```diff
@@ -75,6 +75,10 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 	}
 
 	_, err := c.keycloakAccountClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
+	if err != nil {
+		c.logger.Warn("err", err.Error())
+		return err
```
fperot74: You should return nil in line 97

bsoniam: But err should be nil in that case, no?

fperot74: Yes but writing `return err` let think that the returned value could be different

bsoniam: True, you're right. I will change this

PR #124, opened 2019-09-26
[CLOUDTRUST-1713] record 2ND_FACTOR_REMOVED event

bsoniam:
if a credential other than the password was deleted, record the event 2ND_FACTOR_REMOVED in the audit DB

### DIFF
--- a/pkg/management/component.go
+++ b/pkg/management/component.go
@@ -689,12 +689,12 @@ func (c *component) DeleteCredentialsForUser(ctx context.Context, realmName stri
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 
 	// get the list of credentails of the user
-	credsKc, err := c.keycloakClient.GetCredentialsForUser(accessToken, ctxRealm, realmName, userID)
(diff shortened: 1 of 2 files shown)


pkg/management/component.go
```diff
@@ -688,10 +688,36 @@ func (c *component) GetCredentialsForUser(ctx context.Context, realmName string,
...
+				//store in the logs also the event that failed to be stored in the DB
+				m := map[string]interface{}{"event_name": "2ND_FACTOR_REMOVED", database.CtEventRealmName: realmName, database.CtEventUserID: userID}
+				eventJSON, errMarshal := json.Marshal(m)
+				if errMarshal == nil {
+					c.logger.Error("err", errEvent.Error(), "event", string(eventJSON))
+				} else {
+					c.logger.Error("err", errEvent.Error())
+				}
+			}
+			break
```
fperot74: I think we can split the check `id=credentialID && type=password` in 2 distinct `if` and we can `break` if `id=credentialID` whenever type is "password" or not

(we should not do that if 2 distinct credentials can have the same ID with different types... but it should not happen)

PR #125, opened 2019-09-26
[CLOUDTRUST-1711] Exclude some events from GetEvents

fperot74:
Add a parameter to GetEvents : exclude
It will exclude some ctEvenType values from the request (could be a comma-separated multiple value in the future but not yet implemented)


internal/keycloakb/eventsdbmodule_test.go
```diff
@@ -19,16 +19,26 @@ func TestModuleGetEvents(t *testing.T) {
...
+		var empty [0]api.AuditRepresentation
+		var expectedResult = empty[:]
+		var expectedError error = errorhandler.CreateMissingParameterError("")
+		var rows sql.Rows
+		dbEvents.EXPECT().Query(gomock.Any(), params["origin"], nil, nil, nil, nil, nil, nil, 0, params["max"]).Return(&rows, expectedError).Times(1)
+		res, err := module.GetEvents(context.Background(), params)
+
+		assert.Equal(t, expectedResult, res)
+		assert.Equal(t, expectedError, err)
+	}
```
bsoniam: Would it be possible to do a test where we actually return a list of events but exclude the ones we asked to exclude ?

fperot74: Yes, it would ;-)

fperot74: Sorry. We currently don't know how to write unit tests throught the DB API.

PR #128, opened 2019-10-08
Cloudtrust 1771 fix unittest

harture:
One of the unittests was failing in our local CI.
I think Travis locale is UTC so it was not detected here.


harture:
J'aime pas non plus :(

PR #131, opened 2019-10-16
[CLOUDTRUST-1707] generate TEMPORARILY_LOCKED event when max login failures are exceeded

bsoniam:


pkg/event/component_test.go
```diff
@@ -267,6 +267,26 @@ func TestEventToMapLogonError(t *testing.T) {
...
+		fb.EventStart(builder)
+		fb.EventAddType(builder, etype)
+		fb.EventAddError(builder, error)
+		var eventOffset = fb.EventEnd(builder)
+		builder.Finish(eventOffset)
+		event = fb.GetRootAsEvent(builder.FinishedBytes(), 0)
+	}
+
+	var m = eventToMap(event)
+	assert.Equal(t, "TEMPORARILY_LOCKED", m[database.CtEventType])
```
fperot74: Please add a test with a different error and one with a missing error (to ensure that f["error"] is not failing)

bsoniam: Good suggestion; 
The missing error would correspond to TestEventToMapLogonError
I will add the one with diff error

fperot74: I ran the tests to check coverage: I think the different error is already check

PR #136, opened 2019-10-25
[CLOUDTRUST-1752] correcting the error format + logging errors

bsoniam:


pkg/account/component.go
```diff
@@ -262,8 +262,21 @@ func (c *component) UpdateLabelCredential(ctx context.Context, credentialID stri
 
 	//store the API call into the DB
 	// the error should be treated
-	additionalInfos, _ := json.Marshal(map[string]string{"credentialID": credentialID, "label": label})
-	_ = c.reportEvent(ctx, "SELF_UPDATE_CREDENTIAL", database.CtEventRealmName, currentRealm, database.CtEventUserID, userID, database.CtEventUsername, username, database.CtEventAdditionalInfo, string(additionalInfos))
+	additionalInfos, err := json.Marshal(map[string]string{"credentialID": credentialID, "label": label})
+	if err != nil {
```
fperot74: I think this change is not useful... Do we really think we can get an error here ? (but it is not a problem)

pkg/account/component.go
```diff
@@ -194,7 +187,11 @@ func (c *component) UpdateAccount(ctx context.Context, user api.AccountRepresent
 	}
 
 	//store the API call into the DB
-	_ = c.reportEvent(ctx, "UPDATE_ACCOUNT", database.CtEventRealmName, realm, database.CtEventUserID, userID, database.CtEventUsername, username)
+	errEvent := c.reportEvent(ctx, "UPDATE_ACCOUNT", database.CtEventRealmName, realm, database.CtEventUserID, userID, database.CtEventUsername, username)
```
harture: Maybe we could also have a method for those line as it seems to be very similar every time ? 
What do you think of it ?

bsoniam: DONE

PR #146, opened 2019-12-13
[CLOUDTRUST-1695] Reset SmsSent attribute + add it in the user repr

bsoniam:


pkg/management/component.go
```diff
@@ -633,6 +635,35 @@ func (c *component) SendReminderEmail(ctx context.Context, realmName string, use
...
+	if err != nil {
+		c.logger.Warn(ctx, "err", err.Error())
+		return err
+	}
+
+	//reset the counter, if the smsSent attribute exists
+	resetCounter := 0
+	if userKc.Attributes != nil {
+		var m = *userKc.Attributes
+		if m["smsSent"] != nil {
```
fperot74: As an option, we could avoid calling kc.UpdateUser is m["smsSent"] is nil or equal to 0 (but not an issue)

pkg/management/component_test.go
```diff
@@ -1699,6 +1699,57 @@ func TestSendReminderEmail(t *testing.T) {
...
+
+		err := managementComponent.ResetSmsCounter(ctx, "master", userID)
+
+		assert.Nil(t, err)
+	}
+
+	// Error at GetUser
+	{
+
+		mockKeycloakClient.EXPECT().GetUser(accessToken, realmName, userID).Return(kc.UserRepresentation{}, fmt.Errorf("error")).Times(1)
```
fperot74: You also can write a test when m["smsSent"] is already set (with any value)

bsoniam: Done

pkg/management/component.go
bsoniam: Done

PR #148, opened 2019-12-16
[CLOUDTRUST-2064] Add liveness probe

fperot74:
Implement liveness probe
Feature implemented in cloudtrust/common-service#30
Feature referenced in cloudtrust/keycloak-client#34
Feature used by cloudtrust/keycloak-bridge#148

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,39 @@ It is manadatory to load an authorization JSON file, if no authorization file is
 
 The keycloak event-emitter module sends all events to the bridge's event endpoint. The event emitter use HTTP with flatbuffers.
 
+### Monitoring of keycloak-bridge
+
+An endpoint allows to get a status of the Bridge and its components health.
+URL: ```http://{bridge.host}:{bridge.port}/health/check```
+
+Status example:
+```
+{
+ "name": "keycloak-bridge",
+ "state": "DOWN",
+ "details": [
+  {
+   "name": "Audit R/W",
+   "type": "database",
+   "state": "DOWN",
+   "message": "bad connection"
+  },
+  {
+   "name": "Config RO",
+   "type": "database",
+   "state": "UP",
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -86,6 +86,37 @@ It is manadatory to load an authorization JSON file, if no authorization file is
...
+Status example:
+```
+{
+ "Audit R/W": {
+  "state": "UP",
+  "connection": "established"
+ },
+ "Audit RO": {
+  "state": "DOWN",
+  "message": "commands out of sync. You can't run this command now"
```
bsoniam: In what case we would have this message?

README.md
```diff
@@ -86,6 +86,37 @@ It is manadatory to load an authorization JSON file, if no authorization file is
...
 
+### Monitoring of keycloak-bridge
+
+An endpoint allows to get a status of the Bridge and its components health.
+URL: ```http://{bridge.host}:{bridge.port}/health/check```
+
+Status example:
+```
+{
+ "Audit R/W": {
```
harture: Maybe add something to state it is database connection ?

README.md
fperot74: It is the message sent by the MariaDB driver when you're trying to execute a SQL query while the database has been shut down.

README.md
fperot74: Something like this ?
```
{
 "Audit R/W": {
  "state": "UP",
  "type": "database",
  "connection": "established"
 },
 "Keycloak": {
  "state": "DOWN",
  "type": "http",
  "message": "Can't hit target: Get http://127.0.0.1:8080: dial tcp 127.0.0.1:8080: connectex: No connection could be made because the target machine actively refused it."
 }
```

README.md
bsoniam: Thanks

PR #149, opened 2019-12-20
[CLOUDTRUST-2125] record an event when execute actions emails is called

bsoniam:


pkg/management/component.go
```diff
@@ -589,15 +589,24 @@ func (c *component) ExecuteActionsEmail(ctx context.Context, realmName string, u
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 
 	var actions = []string{}
-
+	var eventRecorded bool = false
```
fperot74: `var eventRecorded = false` should be enough !?

pkg/management/component_test.go
```diff
@@ -1596,6 +1596,18 @@ func TestExecuteActionsEmail(t *testing.T) {
 
 		assert.NotNil(t, err)
 	}
+	// Send email actions, but not sms-password-set
+	{
+
```
fperot74: `var actions2` and `var reqActions2` are only used here... they could be declared here

pkg/management/component.go
```diff
@@ -589,15 +589,24 @@ func (c *component) ExecuteActionsEmail(ctx context.Context, realmName string, u
...
 	for _, requiredAction := range requiredActions {
 		actions = append(actions, string(requiredAction))
 		if string(requiredAction) == initPasswordAction {
 			//store the API call into the DB
 			c.reportEvent(ctx, "INIT_PASSWORD", database.CtEventRealmName, realmName, database.CtEventUserID, userID)
+			eventRecorded = true
 		}
 	}
 
+	if !eventRecorded {
```
harture: With this mechanism, the password init could hide some other required action.
What do you think about record the ACTION_EMAIL in any case ?

pkg/management/component.go
bsoniam: Should be enough, no? Do you see a problem here ?

pkg/management/component_test.go
bsoniam: Thank you for the suggestion
DONE

pkg/management/component.go
bsoniam: Yes, we can do this
DONE

PR #150, opened 2019-12-23
[CLOUDTRUST-2124] validate the max param in GetStatisticsAuthenticationsLog

bsoniam:


pkg/statistics/component.go
```diff
@@ -45,6 +46,10 @@ func NewComponent(db keycloakb.EventsDBModule, keycloakClient KeycloakClient, lo
 	}
 }
 
+const (
+	RegExpMax = `^\d{1,2}$`
```
fperot74: Could be lower case if only used here... but I think it would be better to move this constant in api/statistics/api.go with other RegExp constants... something like RegExpTwoDigitsNumber or RegExpSmallNumber ?

bsoniam: My mistake, this has been corrected

pkg/statistics/component.go
```diff
@@ -139,6 +140,13 @@ func (ec *component) GetStatisticsAuthenticationsLog(ctx context.Context, realmN
 
 	var res []api.StatisticsConnectionRepresentation
 
+	ok, _ := regexp.MatchString(api.RegExpTwoDigitsNumber, max)
+
+	if !ok {
```
fperot74: Go allows you to write this on a single line ;-)
if ok, _ := regexp.MatchString(api.RegExpTwoDigitsNumber, max); !ok {
}

bsoniam: Done

PR #156, opened 2020-01-15
[CLOUDTRUST-2109] Authorization refactoring

harture:
We plan to split the authorizations refactoring in 2 phases (2 PR).
First, we deploy the 1st PR which contains the authorization edition/management. We perform the authorizations configuration which will be stored in DB. This config will not be used, the JSON is still used to check the authorizations.
Second step, we deploy the 2nd PR which will remove the usage of JSOn and compute the authorization from the DB configuration.
This strategy should allow us to minimize the downtime and migration risks.


harture:
We plan to split the authorizations refactoring in 2 phases (2 PR).
First, we deploy the 1st PR which contains the authorization edition/management. We perform the authorizations configuration which will be stored in DB. This config will not be used, the JSON is still used to check the authorizations.
Second step, we deploy the 2nd PR which will remove the usage of JSOn and compute the authorization from the DB configuration.
This strategy should allow us to minimize the downtime and migration risks.

internal/keycloakb/configdbmodule.go
```diff
@@ -67,3 +76,76 @@ func (c *configurationDBModule) GetConfiguration(context context.Context, realmI
 		return config, err
 	}
 }
+
+func (c *configurationDBModule) GetAuthorizations(ctx context.Context, realmID string, groupName string) ([]dto.Authorization, error) {
```
fperot74: May be reused in paper-card-service

internal/keycloakb/configdbmodule_test.go
```diff
@@ -8,25 +8,29 @@ import (
...
+	"github.com/cloudtrust/common-service/log"
 	"github.com/cloudtrust/keycloak-bridge/internal/dto"
 	"github.com/cloudtrust/keycloak-bridge/internal/keycloakb/mock"
+	msg "github.com/cloudtrust/keycloak-bridge/internal/messages"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 )
 
-func useMockDB(t *testing.T, fn func(DBConfiguration)) {
+func useMockDB(t *testing.T, fn func(sqltypes.CloudtrustDB)) {
```
fperot74: This function is not used... it can be removed

pkg/events/component.go
```diff
@@ -40,6 +42,23 @@ func (ec *component) reportEvent(ctx context.Context, apiCall string, values ...
 
 }
 
+// Get actions
+func (ec *component) GetActions(ctx context.Context) ([]api.ActionRepresentation, error) {
```
fperot74: Could be part of the common lib

api/management/api.go
```diff
@@ -242,6 +254,97 @@ func ConvertToKCUser(user UserRepresentation) kc.UserRepresentation {
 	return userRep
 }
 
+// ConvertToKCGroup creates a KC group representation from an API group
+func ConvertToKCGroup(group GroupRepresentation) kc.GroupRepresentation {
+	var groupRep kc.GroupRepresentation
+
+	groupRep.Name = group.Name
+
+	return groupRep
```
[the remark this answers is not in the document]
harture: Thanks, indeed it is better like that ! ;)

internal/keycloakb/configdbmodule.go
harture: When we will retrieve the authorizations to apply authorizations, we will retrieve all of them not just one for a specific realm and group.

internal/keycloakb/configdbmodule_test.go
harture: thanks

pkg/events/component.go
harture: I agree this method is very similar for each Component but a simple and elegant way to do it ...
Would you have any idea ?

fperot74: Let's say we will have a look at this when we'll have time for this... (it almost means never)

api/events/api.go
```diff
@@ -2,6 +2,12 @@ package events_api
 
 import "database/sql"
 
+// ActionRepresentation struct
+type ActionRepresentation struct {
```
bsoniam: The same structure is used by papercard module.  
If for the long term this stays the same, then we can move them in common service

api/events/swagger-api_events.yaml
```diff
@@ -9,6 +9,20 @@ tags:
...
+      summary: Get the list of all possible actions
+      responses:
+        200:
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
```
bsoniam: Is not the same as you code in the paper card?
  /cards/actions:
    get:
      tags:
      - Actions
      summary: Get the list of all possible actions
      responses:
        200:
          content:
            application/json:
              schema:
                type: array
                items:
                 ** $ref: '#/components/schemas/Actions' **

pkg/statistics/authorization.go
```diff
@@ -8,14 +8,27 @@ import (
 	api "github.com/cloudtrust/keycloak-bridge/api/statistics"
 )
 
+var actions []security.Action
+
+func newAction(as string, scope security.Scope) security.Action {
+	a := security.Action{
```
bsoniam: Same

api/events/swagger-api_events.yaml
harture: You are right, it was a bug, it is fixed in the PR #159

pkg/events/authorization.go
```diff
@@ -8,11 +8,24 @@ import (
 	api "github.com/cloudtrust/keycloak-bridge/api/events"
 )
 
+var actions []security.Action
+
+func newAction(as string, scope security.Scope) security.Action {
+	a := security.Action{
```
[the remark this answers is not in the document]
harture: I think I can remove this Id, we finally don't use it.

api/events/api.go
harture: I will move it in common during PR2

api/management/swagger-api_management.yaml
```diff
@@ -15,6 +15,20 @@ tags:
 - name: Roles
   description: Roles management
 paths:
+  /actions:
+    get:
+      tags:
```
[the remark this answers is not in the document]
harture: TODO in PR2, thks

internal/messages/errormessages.go
```diff
@@ -33,8 +34,10 @@ const (
 	PhoneNumber        = "phoneNumber"
 	Gender             = "gender"
 	Birthdate          = "birthdate"
-	GroudID            = "groupId"
-	GroudIDs           = "groupIds"
+	GroupID            = "groupId"
+	GroupName          = "groupName"
+	Name               = "name"
```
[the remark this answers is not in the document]
harture: I will fix it

PR #160, opened 2020-01-24
[CLOUDTRUST-2208] change math/rand with crypto/rand

bsoniam:


internal/keycloakb/passwordgeneration.go
```diff
@@ -19,7 +21,9 @@ const (
 func appendCharacters(pwdElems []string, alphabet string, length int) []string {
 
 	for j := 0; j < length; j++ {
-		pwdElems = append(pwdElems, string(alphabet[rand.Intn(len(alphabet))]))
+		nBig, _ := crand.Int(crand.Reader, big.NewInt(int64(len(alphabet))))
```
fperot74: You could evaluate `big.NewInt(int64(len(alphabet)))` in a local variable `maxInt` outside of the loop to make it more readable (and very very little bit more performant)

internal/keycloakb/passwordgeneration.go
```diff
@@ -102,7 +106,7 @@ func GeneratePasswordFromKeycloakPolicy(policy string) (string, error) {
 			}
 		}
 	}
-	rand.Shuffle(len(pwdElems), func(i, j int) { pwdElems[i], pwdElems[j] = pwdElems[j], pwdElems[i] })
+	mrand.Shuffle(len(pwdElems), func(i, j int) { pwdElems[i], pwdElems[j] = pwdElems[j], pwdElems[i] })
```
harture: this shuffle is not using the cryptographic random source but not sure if it really matter ?

bsoniam: it should not matter

PR #161, opened 2020-01-24
[CLOUDTRUST-2107] Recaptcha processing for register user

fperot74:


pkg/register/authorization.go
```diff
@@ -18,8 +17,7 @@ import (
 )
 
 const (
-	regexpBasicAuth = `^[Bb]asic (.+)$`
-	regExpRecaptcha = `^([\w\d]+):secret=(.+),token=(.+)$`
+	regexpRecaptchaToken = `^[\w\d-]+$`
```
harture: Do we have a max size for this token ?

fperot74: We had an example with almost 480 characters but we don't really have details about this token.

PR #163, opened 2020-01-27
CLOUDTRUST-2200: add API for setting the trustID groups

lagess:


api/management/api.go
```diff
@@ -186,6 +187,11 @@ func ConvertToAPIUser(userKc kc.UserRepresentation) UserRepresentation {
 			counter, _ := strconv.Atoi(smsSent)
 			userRep.SmsSent = &counter
 		}
+
+		if m["trustid_groups"] != nil {
```
bsoniam: I would suggest trustIDGroups (to respect the same format as the others)

api/management/api_test.go
```diff
@@ -74,6 +74,12 @@ func TestConvertToAPIUser(t *testing.T) {
 	kcUser.Attributes = &m
 	m["smsSent"] = []string{"0"}
 	assert.NotNil(t, *ConvertToAPIUser(kcUser).SmsSent)
+	
+	// trustID groups
+	assert.Nil(t, ConvertToAPIUser(kcUser).TrustIDGroups)
+	kcUser.Attributes = &m
+	m["trustid_groups"] = []string{"en"}
```
bsoniam: same

internal/messages/errormessages.go
```diff
@@ -39,6 +39,7 @@ const (
 	GroupName          = "groupName"
 	Name               = "name"
 	GroupIDs           = "groupIds"
```
bsoniam: it's interesting that we have now singular and plural forms for several nouns

api/management/api.go
lagess: Done :-)

api/management/api_test.go
lagess: Fixed

internal/messages/errormessages.go
lagess: GropuNames removed

internal/messages/errormessages.go
```diff
@@ -61,4 +61,5 @@ const (
 	Unit               = "unit"
 	Max                = "max"
 	Timeshift          = "timeshift"
+	TrustIDGroupName   = "TrustIDGroupName"
```
fperot74: This parameter is the only one to start with an uppercase letter?

pkg/management/component.go
```diff
@@ -116,20 +117,22 @@ type Component interface {
...
 // Component is the management component.
 type component struct {
-	keycloakClient KeycloakClient
-	eventDBModule  database.EventsDBModule
-	configDBModule ConfigurationDBModule
-	logger         keycloakb.Logger
+	keycloakClient          KeycloakClient
+	eventDBModule           database.EventsDBModule
+	configDBModule          ConfigurationDBModule
+	authorizedTrustIDGroups []string
```
fperot74: When instanciating this struct, you could convert the `[]string` into `map[string]bool`...
In SetTrustIDGroups, you would only have to check like this : `if _, ok := c.authorizedTrustIDGroups[groupName]; ok {...}` and avoid to have to check all values of your array (it would also make SetTrustIDGroups be shorter and so, easier to read)

pkg/management/component.go
```diff
@@ -495,6 +498,49 @@ func (c *component) GetGroupsOfUser(ctx context.Context, realmName, userID strin
...
 }
 
+func (c *component) SetTrustIDGroups(ctx context.Context, realmName, userID string, groupNames []string) error {
+	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
+
+	// validate the input - trustID groups must be valid
+	var extGroupNames []string
+	for _, groupName := range groupNames {
+		var match = false
+		for _, authzedGrpName := range c.authorizedTrustIDGroups {
```
fperot74: You can avoid this whole `for` statement by using a map as previously explained

pkg/management/component.go
```diff
@@ -495,6 +498,49 @@ func (c *component) GetGroupsOfUser(ctx context.Context, realmName, userID strin
...
+			if groupName == authzedGrpName {
+				extGroupNames = append(extGroupNames, "/"+groupName)
+				match = true
+				break
+			}
+		}
+		if !match {
```
fperot74: With map[string]bool, this line would be: `if _, ok := c.authorizedTrustIDGroups[groupName]; ok {`

pkg/management/component_test.go
```diff
@@ -1382,15 +1401,105 @@ func TestGetGroupsOfUser(t *testing.T) {
...
+	var managementComponent = NewComponent(mockKeycloakClient, mockEventDBModule, mockConfigurationDBModule, allowedTrustIDGroups, mockLogger)
+
+	var accessToken = "TOKEN=="
+
+	var username = "user"
+	var realmName = "master"
+	var userID = "789-1234-5678"
+
+	// Set groups with success
+	{
```
fperot74: You can use t.Run(...) (something we recently discovered :-) )
```
t.Run("Set groups with success", func(t *testing.T) {
    // Your test
})
```

internal/messages/errormessages.go
lagess: fixed

pkg/management/component.go
lagess: Done

pkg/management/component.go
lagess: done

pkg/management/component.go
lagess: I've put else :-)

pkg/management/component_test.go
lagess: Good to know, fixed!

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -215,6 +215,13 @@ func main() {
 		}
 	}
 
+	// Security - allowed trustID groups
+	var trustIDGroups = strings.Split(c.GetString("trustid-groups"), ",")
```
fperot74: You can use c.GetStringSlice("trustid-groups") and have them as multiple values in the config file (see cors config)

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -215,6 +215,13 @@ func main() {
...
+	var authzedTrustIDGroups = make(map[string]bool)
+	for _, grp := range trustIDGroups {
+		authzedTrustIDGroups[grp] = true
```
fperot74: I would have write this in the NewComponent(...) function... but it works too

lagess: fixed

cmd/keycloakb/keycloak_bridge.go
lagess: fixed

PR #171, opened 2020-01-29
[CLOUDTRUST-2252] Success registration url from account

harture:


fperot74:
Coverage -0,02% ! I wonder if an approval is well deserved... ;-)

PR #172, opened 2020-01-30
fixup regexp

harture:


bsoniam:
-

pkg/account/http.go
```diff
@@ -28,7 +28,7 @@ func decodeAccountRequest(ctx context.Context, req *http.Request) (interface{},
 	}
 
 	var queryParams = map[string]string{
-		"realm_id": account_api.RegExpID,
+		"realm_id": account_api.RegExpRealmName,
```
fperot74: Awful ! Did you plan to update the parameter name later to avoid having id=name?

PR #174, opened 2020-01-30
[CLOUDTRUST-2254] fix SQL query for retrieving events

lagess:


lagess:
@fperot74 You are right... Thanks for pointing out!

internal/keycloakb/eventsdbmodule.go
```diff
@@ -54,12 +54,12 @@ type selectAuditEventsParameters struct {
...
-	AND realm_name = IFNULL(?, realm_name)
-	AND user_id = IFNULL(?, user_id)
-	AND ct_event_type = IFNULL(?, ct_event_type)
+	WHERE (? IS NULL OR origin = ?)
+	AND (? IS NULL OR realm_name = ?)
+	AND (? IS NULL OR user_id = ?)
+	AND (? IS NULL OR ct_event_type = ?)
 	AND unix_timestamp(audit_time) between IFNULL(?, unix_timestamp(audit_time)) and IFNULL(?, unix_timestamp(audit_time))
-	AND ct_event_type <> IFNULL(?, 'not-a-ct-event-type')
+	AND (? IS NULL OR ct_event_type <> ?)
```
fperot74: Change was not necessary as ct_event_type is not supposed to be null

lagess: Better be safe that sorry :-) We never know what we will implement in the future...

PR #176, opened 2020-01-31
[CLOUDTRUST-2264] Add GetUser (by id) in kyc API

fperot74:
Coverage should decrease but I don't want to write unit tests for deprecated methods that will be removed soon.

### DIFF
--- a/api/kyc/swagger-api_kyc.yaml
+++ b/api/kyc/swagger-api_kyc.yaml
@@ -31,13 +31,19 @@ paths:
     get:
       tags:
       - KYC
-      summary: Gets a user
+      summary: Gets a user (search by username)
       security:
         - openId: []
       parameters:
       - name: username
         in: query
         description: name of the user to be retrieved
+        required: true
+        schema:
+          type: string
+      - name: groupIds
+        in: query
+        description: list of groupId the users may belong to (list comma seperated)
         required: true
         schema:
           type: string
@@ -50,9 +56,29 @@ paths:
                 $ref: '#/components/schemas/User'
         403:
           description: No permission to call this operation
-        404:
-          description: User not found
   /kyc/users/{userId}:
+    get:
+      tags:
+      - KYC
+      summary: Gets a user
+      security:
+        - openId: []
+      parameters:
+      - name: userId
+        in: path
+        description: user id
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/User'
+        403:
+          description: No permission to call this operation
     put:
       tags:
       - KYC
@@ -75,7 +101,7 @@ paths:
               $ref: '#/components/schemas/User'
       responses:
         200:
-          description: Successful operation. Returns the generated username
+          description: Successful operation
           content:
             application/json:
               schema:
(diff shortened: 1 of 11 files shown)


api/kyc/swagger-api_kyc.yaml
```diff
@@ -50,9 +50,29 @@ paths:
...
+          type: string
+      responses:
+        200:
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/User'
+        403:
+          description: No permission to call this operation
```
bsoniam: There is no more 404 possible?

fperot74: I'ld like so but Keycloak doesn't return 404 when users are not found

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -631,9 +631,10 @@ func main() {
 		kycComponent = kyc.MakeAuthorizationRegisterComponentMW(registerRealm, log.With(kycLogger, "mw", "endpoint"), authorizationManager)(kycComponent)
 
 		kycEndpoints = kyc.Endpoints{
-			GetActions:   prepareEndpoint(kyc.MakeGetActionsEndpoint(kycComponent), "register_get_actions", influxMetrics, kycLogger, tracer, rateLimit["kyc"]),
-			GetUser:      prepareEndpoint(kyc.MakeGetUserEndpoint(kycComponent), "get_user", influxMetrics, kycLogger, tracer, rateLimit["kyc"]),
-			ValidateUser: prepareEndpoint(kyc.MakeValidateUserEndpoint(kycComponent), "validate_user", influxMetrics, kycLogger, tracer, rateLimit["kyc"]),
+			GetActions:    prepareEndpoint(kyc.MakeGetActionsEndpoint(kycComponent), "register_get_actions", influxMetrics, kycLogger, tracer, rateLimit["kyc"]),
+			GetUser:       prepareEndpoint(kyc.MakeGetUserEndpoint(kycComponent), "get_user", influxMetrics, kycLogger, tracer, rateLimit["kyc"]),
+			GetUserByName: prepareEndpoint(kyc.MakeGetUserByNameEndpoint(kycComponent), "get_user_by_name", influxMetrics, kycLogger, tracer, rateLimit["kyc"]),
```
harture: GetUserByUsername ?

pkg/kyc/authorization.go
```diff
@@ -63,7 +63,7 @@ func (c *authorizationComponentMW) GetActions(ctx context.Context) ([]apikyc.Act
 	return c.next.GetActions(ctx)
 }
 
-func (c *authorizationComponentMW) GetUser(ctx context.Context, username string) (apikyc.UserRepresentation, error) {
+func (c *authorizationComponentMW) GetUserByName(ctx context.Context, username string) (apikyc.UserRepresentation, error) {
 	var action = KYCGetUser.String()
 	var targetRealm = c.realmName
 	var groupID = RegistrationOfficer
```
harture: We should call CheckAuthorizationOnTargetUser. 
groupID is the one of the target so it seems incorrect there.

pkg/kyc/authorization.go
```diff
@@ -72,7 +72,19 @@ func (c *authorizationComponentMW) GetUser(ctx context.Context, username string)
...
-	return c.next.GetUser(ctx, username)
+	return c.next.GetUserByName(ctx, username)
+}
+
+func (c *authorizationComponentMW) GetUser(ctx context.Context, userID string) (apikyc.UserRepresentation, error) {
+	var action = KYCGetUser.String()
+	var targetRealm = c.realmName
+	var groupID = RegistrationOfficer
+
+	if err := c.authManager.CheckAuthorizationOnTargetGroup(ctx, action, targetRealm, groupID); err != nil {
```
harture: Same ass before

cmd/keycloakb/keycloak_bridge.go
fperot74: Done

PR #179, opened 2020-02-06
Add a parameter validation tool

fperot74:
Suggestion to have tools to make check of parameters easier.
Could be added in cloudtrust/common-service


fperot74:
Feature moved to cloudtrust/common-service#37

PR #180, opened 2020-02-10
[CLOUDTRUST-2258] Set the locale when the user registers himself

fperot74:
+ uses parameters validation from common-service

⚠️ Note that validation code is much more shorter now so the decrease of coverage is something that could be expected (less code / same features)


api/account/api.go
```diff
@@ -107,6 +111,9 @@ func ConvertToKCUser(user AccountRepresentation) kc.UserRepresentation {
 	if user.PhoneNumber != nil {
 		attributes["phoneNumber"] = []string{*user.PhoneNumber}
 	}
+	if user.Locale != nil {
+		attributes["locale"] = []string{*user.Locale}
```
harture: I cannot directly put this comment on the right line (128), I think the validation of the locale is missing in the Validate method

fratt-elca:
Merge this at the same time :
https://git.elcanet.local/projects/CLOUDTRUST/repos/2019.oc.register/pull-requests/7

api/kyc/api.go
```diff
@@ -157,46 +157,16 @@ func (u *UserRepresentation) ImportFromKeycloak(kcUser *kc.UserRepresentation) {
...
+		ValidateParameterIn(prmUserGender, u.Gender, allowedGender, true).
+		ValidateParameterRegExp(prmUserFirstName, u.FirstName, regExpFirstName, true).
+		ValidateParameterRegExp(prmUserLastName, u.LastName, regExpLastName, true).
+		ValidateParameterRegExp(prmUserEmail, u.EmailAddress, regExpEmail, true).
+		ValidateParameterPhoneNumber(prmUserPhoneNumber, u.PhoneNumber, true).
+		ValidateParameterDate(prmUserBirthDate, u.BirthDate, dateLayout, true).
+		ValidateParameterRegExp(prmUserBirthLocation, u.BirthLocation, regExpBirthLocation, true).
+		ValidateParameterIn(prmUserIDDocumentType, u.IDDocumentType, allowedDocumentType, true).
+		ValidateParameterRegExp(prmUserIDDocumentNumber, u.IDDocumentNumber, regExpIDDocumentNumber, true).
+		ValidateParameterDate(prmUserIDDocumentExpiration, u.IDDocumentExpiration, dateLayout, true).
```
bsoniam: we have no locale here?

internal/messages/errormessages.go
```diff
@@ -33,6 +33,7 @@ const (
 	Email                             = "email"
 	Firstname                         = "firstname"
 	Lastname                          = "lastname"
+	Locale                            = "locale"
```
bsoniam: improving the proximity factor :)

fperot74: Not really... I added a "new" constant... but it could not be built because it was already present so I removed one but not removed mine... I'll revert this change

api/kyc/api.go
fperot74: We could... but is it the mission of the validation operator to decide which is the user's preffered language?

PR #183, opened 2020-02-19
[CLOUDTRUST-2295] Bridge endpoint to self-ask for an email validation

fperot74:
[CLOUDTRUST-2296] Bridge endpoint to self-ask for a phone number validation
[CLOUDTRUST-2299] Send email when user updates his email/phone. Unvalidate his account validation

Refers to:
* cloudtrust/keycloak-rest-api-extensions#26
* cloudtrust/keycloak-client#46


Gopkg.toml
```diff
@@ -27,11 +27,11 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/common-service"
-  branch = "master"
+  version = "v1.2.5"
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  branch = "dev"
+  branch = "jira-2296"
```
fperot74: To be updated before merging

pkg/account/component.go
```diff
@@ -15,6 +16,12 @@ import (
 	kc "github.com/cloudtrust/keycloak-client"
 )
 
+// Constants
+const (
+	ActionVerifyEmail       = "VERIFY_EMAIL"
```
bsoniam: why 2 diff styles?

pkg/account/component.go
```diff
@@ -27,6 +34,8 @@ type KeycloakAccountClient interface {
 	UpdateAccount(accessToken, realm string, user kc.UserRepresentation) error
 	GetAccount(accessToken, realm string) (kc.UserRepresentation, error)
 	DeleteAccount(accessToken, realm string) error
+	ExecuteActionsEmail(accessToken string, realmName string, actions []string) error
+	//ExecuteActionsEmail(accessToken string, actions []string, paramKV ...string) error
```
bsoniam: maybe remove the second line

pkg/account/component.go
fperot74: Yes... one comes from Keycloak, the other one comes from 2019.kc-sms

api/account/swagger-api_account.yaml
```diff
@@ -160,7 +160,22 @@ paths:
       responses:
         200:
           description: Successful operation.
-
+  /account/send-verify-email:
```
harture: Maybe we could remove the "send" in URL to have /account/verify-email as it would sounds more RESTful

pkg/account/component.go
```diff
@@ -370,3 +387,41 @@ func (c *component) GetConfiguration(ctx context.Context, realmIDOverride string
...
+	if err != nil {
+		c.logger.Warn(ctx, "err", err.Error())
+		return err
+	}
+
+	if contains(actions, ActionVerifyEmail) {
+		var additionalInfos = map[string]string{"actions": strings.Join(actions, ",")}
+		var additionalBytes, _ = json.Marshal(additionalInfos)
+		var additionalString = string(additionalBytes)
+		c.reportEvent(ctx, "ACTION_EMAIL", database.CtEventRealmName, currentRealm, database.CtEventUserID, userID, database.CtEventAdditionalInfo, additionalString)
```
harture: ACTION_EMAIL must be sent only for VERIFY_EMAIL ? If yes, maybe we should have another specific event for verify phone number ?

PR #185, opened 2020-02-21
[CLOUDTRUST-2298] Add fields to realm configuration

fperot74:
Requires:
* cloudtrust/common-service#38


pkg/account/authorization.go
```diff
@@ -141,18 +141,26 @@ func (c *authorizationComponentMW) UpdateAccount(ctx context.Context, account ap
 		return err
 	}
 
+	if !isEnabled(config.APISelfAccountEditingEnabled) {
```
harture: I don't see the difference of behavior between APISelfAccountEditingEnabled and APISelfMailEditingEnabled. Both needs to be enabled to be able to perform an edition. But it is not possible to allow to only edit the email. Should we enhance our mechanism to be able to provide finer grain of authorization for edition or remove the falg related to mail edition ?

fperot74: Removing APISelfMailEditingEnabled

fperot74:
> problem though : The email sent when changing the phone number doesn't let me va

I just retested this. I clicked on the link I received by mail. Received an SMS code and enter it to succesfully verify my phone number

pkg/management/component.go
```diff
@@ -1178,6 +1178,7 @@ func (c *component) GetRealmCustomConfiguration(ctx context.Context, realmName s
 				APISelfAccountDeletionEnabled:       &falseBool,
 				ShowAuthenticatorsTab:               &falseBool,
 				ShowPasswordTab:                     &falseBool,
+				ShowProfileTab:                      &falseBool,
 				ShowMailEditing:                     &falseBool,
```
harture: ShowMailEditing should not exist anymore

pkg/management/component.go
```diff
@@ -1199,6 +1200,7 @@ func (c *component) GetRealmCustomConfiguration(ctx context.Context, realmName s
 		APISelfAccountDeletionEnabled:       config.APISelfAccountDeletionEnabled,
 		ShowAuthenticatorsTab:               config.ShowAuthenticatorsTab,
 		ShowPasswordTab:                     config.ShowPasswordTab,
+		ShowProfileTab:                      config.ShowProfileTab,
 		ShowMailEditing:                     config.ShowMailEditing,
```
harture: same here

pkg/management/component.go
```diff
@@ -1270,6 +1272,7 @@ func (c *component) UpdateRealmCustomConfiguration(ctx context.Context, realmNam
 		APISelfAccountDeletionEnabled:       customConfig.APISelfAccountDeletionEnabled,
 		ShowAuthenticatorsTab:               customConfig.ShowAuthenticatorsTab,
 		ShowPasswordTab:                     customConfig.ShowPasswordTab,
+		ShowProfileTab:                      customConfig.ShowProfileTab,
 		ShowMailEditing:                     customConfig.ShowMailEditing,
```
harture: same here

pkg/management/component.go
fperot74: Removed

PR #188, opened 2020-02-25
[CLOUDTRUST-2294] Add endpoint to manage back-office configuration

fperot74:


api/management/api.go
```diff
@@ -380,6 +394,23 @@ func ConvertToKCFedID(fedID FederatedIdentityRepresentation) kc.FederatedIdentit
 
 // Validators
 
+// NewBackOfficeConfigurationFromJSON creates and validate a new BackOfficeConfiguration from a JSON value
```
bsoniam: validates

api/management/swagger-api_management.yaml
```diff
@@ -968,6 +968,77 @@ paths:
           description: successful operation
         400:
           description: invalid information provided  (invalid client identifier or redirect URI not allowed for this client)
+  /realms/{realm}/backoffice-configuration:
+    get:
+      tags:
+      - Configuration
+      summary: Get the current backoffice configuration. Uses groups of the current user.
```
bsoniam: this one uses also the groups of the user? the two endpoints do the same thing?

fperot74: I updated the description... One "get" method is relative to the current user (its realm and its groups)... The two other methods are used to get and set the back-office configuration for a given realmID/groupName

fratt-elca: Please document what happens if a user has multiple groups :-)

internal/keycloakb/configdbmodule.go
```diff
@@ -68,6 +119,58 @@ func (c *configurationDBModule) GetConfiguration(ctx context.Context, realmID st
 	return config, err
 }
 
+func (c *configurationDBModule) GetBackOfficeConfiguration(ctx context.Context, realmID string, groupNames []string) (dto.BackOfficeConfiguration, error) {
+	var res = make(dto.BackOfficeConfiguration)
+	for _, slice := range splitIntoSlices(groupNames, 5) {
```
harture: Maybe we could do something like that: https://stackoverflow.com/questions/20271123/how-to-execute-an-in-lookup-in-sql-using-golang ?

pkg/management/authorization.go
```diff
@@ -22,43 +22,46 @@ func newAction(as string, scope security.Scope) security.Action {
...
+	MGMTCreateGroup                         = newAction("MGMT_CreateGroup", security.ScopeRealm)
+	MGMTDeleteGroup                         = newAction("MGMT_DeleteGroup", security.ScopeGroup)
+	MGMTGetAuthorizations                   = newAction("MGMT_GetAuthorizations", security.ScopeGroup)
+	MGMTUpdateAuthorizations                = newAction("MGMT_UpdateAuthorizations", security.ScopeGroup)
+	MGMTGetClientRoles                      = newAction("MGMT_GetClientRoles", security.ScopeRealm)
+	MGMTCreateClientRole                    = newAction("MGMT_CreateClientRole", security.ScopeRealm)
+	MGMTGetRealmCustomConfiguration         = newAction("MGMT_GetRealmCustomConfiguration", security.ScopeRealm)
+	MGMTUpdateRealmCustomConfiguration      = newAction("MGMT_UpdateRealmCustomConfiguration", security.ScopeRealm)
+	MGMTGetRealmBackOfficeConfiguration     = newAction("MGMT_GetRealmBackOfficeConfiguration", security.ScopeRealm)
+	MGMTUpdateRealmBackOfficeConfiguration  = newAction("MGMTUpdateRealmBackOfficeConfiguration", security.ScopeRealm)
```
harture: Missing _

pkg/management/authorization.go
```diff
@@ -478,6 +481,39 @@ func (c *authorizationComponentMW) UpdateRealmCustomConfiguration(ctx context.Co
 	return c.next.UpdateRealmCustomConfiguration(ctx, realmName, customConfig)
 }
 
+func (c *authorizationComponentMW) GetRealmBackOfficeConfiguration(ctx context.Context, realmName string, groupID string) (api.BackOfficeConfiguration, error) {
+	var action = MGMTGetRealmBackOfficeConfiguration.String()
+	var targetRealm = realmName
+
+	if err := c.authManager.CheckAuthorizationOnTargetRealm(ctx, action, targetRealm); err != nil {
```
harture: As we provide a groupID, I think we could have a CheckAuthorizationOnTargetGroupID
If we change it, we should not forget to adapt scope in the Action definition

pkg/management/authorization.go
```diff
@@ -478,6 +481,39 @@ func (c *authorizationComponentMW) UpdateRealmCustomConfiguration(ctx context.Co
...
+	}
+
+	return c.next.GetRealmBackOfficeConfiguration(ctx, realmName, groupID)
+}
+
+func (c *authorizationComponentMW) UpdateRealmBackOfficeConfiguration(ctx context.Context, realmName string, groupID string, config api.BackOfficeConfiguration) error {
+	var action = MGMTUpdateRealmBackOfficeConfiguration.String()
+	var targetRealm = realmName
+
+	if err := c.authManager.CheckAuthorizationOnTargetRealm(ctx, action, targetRealm); err != nil {
```
harture: Same as before

PR #190, opened 2020-02-28
[CLOUDTRUST-2320] Add endpoints for admin configuration

fperot74:


api/management/swagger-api_management.yaml
```diff
@@ -968,6 +968,46 @@ paths:
           description: successful operation
         400:
           description: invalid information provided  (invalid client identifier or redirect URI not allowed for this client)
+  /realms/{realm}/admin-configuration:
+    get:
+      tags:
+      - Configuration
+      summary: Get the current admin configuration
```
harture: of the current or the one specified in param ?

fperot74: Updating: "Get the admin configuration of the specified realm"

PR #194, opened 2020-03-05
[CLOUDTRUST-2297] Add flag for user accreditations

fperot74:
User accreditations are updated in /kyc and /validation
Accreditations are sent through the "GetUser" API of /kyc, /management and /users

### DIFF
--- a/api/management/api.go
+++ b/api/management/api.go
@@ -2,36 +2,44 @@ package management_api
 
 import (
 	"encoding/json"
-	"strconv"
 
 	errorhandler "github.com/cloudtrust/common-service/errors"
 
 	"github.com/cloudtrust/common-service/configuration"
 	"github.com/cloudtrust/common-service/validation"
 	"github.com/cloudtrust/keycloak-bridge/internal/constants"
+	"github.com/cloudtrust/keycloak-bridge/internal/keycloakb"
 	kc "github.com/cloudtrust/keycloak-client"
 )
 
 // UserRepresentation struct
 type UserRepresentation struct {
-	ID                  *string   `json:"id,omitempty"`
-	Username            *string   `json:"username,omitempty"`
-	Email               *string   `json:"email,omitempty"`
-	Enabled             *bool     `json:"enabled,omitempty"`
-	EmailVerified       *bool     `json:"emailVerified,omitempty"`
-	PhoneNumberVerified *bool     `json:"phoneNumberVerified,omitempty"`
-	FirstName           *string   `json:"firstName,omitempty"`
-	LastName            *string   `json:"lastName,omitempty"`
-	PhoneNumber         *string   `json:"phoneNumber,omitempty"`
-	Label               *string   `json:"label,omitempty"`
-	Gender              *string   `json:"gender,omitempty"`
-	BirthDate           *string   `json:"birthDate,omitempty"`
-	CreatedTimestamp    *int64    `json:"createdTimestamp,omitempty"`
-	Groups              *[]string `json:"groups,omitempty"`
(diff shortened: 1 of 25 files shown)


internal/keycloakb/accreditationsmodule.go
```diff
@@ -0,0 +1,158 @@
...
+	err = am.keycloakClient.UpdateUser(accessToken, realmName, userID, kcUser)
+	if err != nil {
+		am.logger.Warn(ctx, "msg", "CreateAccreditations: can't update Keycloak user", "err", err.Error(), "realm", realmName, "user", userID)
+		return err
+	}
+
+	return nil
+}
+
+func arrayContains(slice []string, value string) bool {
```
bsoniam: used again here

fperot74: I removed this function as validation.IsStringInSlice(...) is now merged in common-service

internal/keycloakb/datehelpers.go
```diff
@@ -46,3 +47,87 @@ func NextMonth(ref time.Time) time.Time {
 	}
 	return time.Date(year, month, 1, 0, 0, 0, 0, ref.Location()).UTC()
 }
+
+type longDuration struct {
```
harture: All those changes are part of another PR in common-service

api/management/api.go
```diff
@@ -525,9 +525,9 @@ func (config RealmCustomConfiguration) Validate() error {
 }
 
 // Validate is a validator for RealmAdminConfiguration
-func (config RealmAdminConfiguration) Validate() error {
+func (rac *RealmAdminConfiguration) Validate() error {
```
harture: Pointeur ?

internal/keycloakb/datehelpers_test.go
```diff
@@ -106,3 +106,40 @@ func TestNextMonth(t *testing.T) {
 	assert.Equal(t, nextMonthIndia, NextMonth(reference.In(locIndia)))
 	assert.Equal(t, nextMonthCookIsland, NextMonth(reference.In(locCookIsland)))
 }
+
+func TestLongDuration(t *testing.T) {
```
harture: those helpers have been moved in common

pkg/validation/component.go
```diff
@@ -212,9 +230,24 @@ func needDBUserUpdate(user api.UserRepresentation) bool {
 }
 
 func (c *component) CreateCheck(ctx context.Context, userID string, check api.CheckRepresentation) error {
+	var accessToken string
+	var err error
+
+	if check.IsIdentificationSuccessful() {
```
harture: I think we should prevent a scenario where we have an accreditation without persisting the check. Maybe we should swap between persistency of accrediation and check ? WHat do you think about it ?

internal/keycloakb/accreditationsmodule.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (am *accredsModule) CreateAccreditations(ctx context.Context, accessToken, realmName, userID, condition string) error {
+	var kcUser, count, err = am.GetUserAndPrepareAccreditations(ctx, accessToken, realmName, userID, condition)
+	if err != nil || count == 0 {
+		return err
+	}
+
+	// Update user in Keycloak
+	err = am.keycloakClient.UpdateUser(accessToken, realmName, userID, kcUser)
```
harture: Is it really the matter of this module to update the user ?
Maybe this accreditations component could only be responsible to create the array of accreditations to persists into the user and let the upper layer (component) to update the user

PR #202, opened 2020-03-19
Fix Swagger (bool used instead of boolean)

fperot74:


fperot74:
Done in cloudtrust/keycloak-bridge#203

PR #203, opened 2020-03-20
[CLOUDTRUST-2435 2436] trustID groups management

fperot74:


api/management/swagger-api_management.yaml
```diff
@@ -115,6 +115,29 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Client'
+  /realms/{realm}/trustIdGroups:
```
bsoniam: In the JIRA the path is  /realms/{realm}/users/trustIdGroups
If it was re-discussed to change it to this path, ignore my comment

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -830,6 +832,8 @@ func main() {
 		var getRolesForUserHandler = configureManagementHandler(keycloakb.ComponentName, ComponentID, idGenerator, keycloakClient, audienceRequired, tracer, logger)(managementEndpoints.GetRolesOfUser)
 		var getGroupsForUserHandler = configureManagementHandler(keycloakb.ComponentName, ComponentID, idGenerator, keycloakClient, audienceRequired, tracer, logger)(managementEndpoints.GetGroupsOfUser)
 		var getUserAccountStatusHandler = configureManagementHandler(keycloakb.ComponentName, ComponentID, idGenerator, keycloakClient, audienceRequired, tracer, logger)(managementEndpoints.GetUserAccountStatus)
+		var getAvailableTrustIDGroupsHandler = configureManagementHandler(keycloakb.ComponentName, ComponentID, idGenerator, keycloakClient, audienceRequired, tracer, logger)(managementEndpoints.GetAvailableTrustIDGroups)
+		var getTrustIDGroupsHandler = configureManagementHandler(keycloakb.ComponentName, ComponentID, idGenerator, keycloakClient, audienceRequired, tracer, logger)(managementEndpoints.GetTrustIDGroups)
```
bsoniam: why not getUserTrustIDGroupsHandler?
or getTrustIDGroupsForUserHandler?
This applies to all the future naming uses

pkg/management/component.go
```diff
@@ -506,6 +508,32 @@ func (c *component) GetGroupsOfUser(ctx context.Context, realmName, userID strin
...
+func (c *component) GetTrustIDGroups(ctx context.Context, realmName, userID string) ([]string, error) {
+	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
+	var currentUser, err = c.keycloakClient.GetUser(accessToken, realmName, userID)
+	if err != nil {
+		c.logger.Warn(ctx, "err", err.Error())
+		return nil, err
+	}
+	var groups = make([]string, 0)
+	for _, grp := range currentUser.GetAttribute(constants.AttrbTrustIDGroups) {
+		if strings.HasPrefix(grp, "/") {
```
bsoniam: out of curiosity: some groups are written with / as preffix ?

pkg/management/component.go
```diff
@@ -506,6 +508,32 @@ func (c *component) GetGroupsOfUser(ctx context.Context, realmName, userID strin
...
+			grp = grp[1:]
+		}
+		groups = append(groups, grp)
```
bsoniam: can it be that one user group is not in the list of c.authorizedTrustIDGroups ?

api/management/swagger-api_management.yaml
fperot74: We had to change the path in order to avoid conflict with URL of GET /realm/{realm}/users/{userID} (trustIdGroups would not be a valid userID but we preffer to avoid this conflict)
I just forgot to update the JIRA description

cmd/keycloakb/keycloak_bridge.go
fperot74: I used the model of existing methods. trustID groups are only referenced at user level

pkg/management/component.go
fperot74: In fact, you should always have a leading /. If you look at method SetTrustIDGroups, you'll see that we always add a / as a prefix. I'm checking if the prefix is here in case of bad values in the attributes but we could potentially remove this test.
(that is also the default behavior of Keycloak user groups)

pkg/management/component.go
fperot74: We spoke of this with @harture and decided to ignore this case. Back office will be able to display a warning

api/management/swagger-api_management.yaml
bsoniam: OK

pkg/management/component.go
bsoniam: thanks for the explanation

cmd/keycloakb/keycloak_bridge.go
bsoniam: on my side I keep my point of view, but alright

pkg/management/component.go
bsoniam: if BO will do this, then maybe we can be a bit ignorant here :)

PR #204, opened 2020-03-20
[CLOUDTRUST-2438] Remove parameter groupIDs from GET /kyc/users

fperot74:


pkg/kyc/authorization.go
```diff
@@ -68,17 +68,15 @@ func (c *authorizationComponentMW) GetActions(ctx context.Context) ([]apikyc.Act
 	return c.next.GetActions(ctx)
 }
 
-func (c *authorizationComponentMW) GetUserByUsername(ctx context.Context, username string, groupIDs []string) (apikyc.UserRepresentation, error) {
+func (c *authorizationComponentMW) GetUserByUsername(ctx context.Context, username string) (apikyc.UserRepresentation, error) {
 	var action = KYCGetUserByUsername.String()
 	var targetRealm = c.realmName
```
harture: We are using realmName of the MW coming from configuration. I think it should depend of the current user and thus using the value present in context. Am I missing something ?

fperot74: Right... fixed

harture: This issue seems to also applies to the others methods

fperot74: Re-fixed

PR #205, opened 2020-03-23
[CLOUDTRUST-2402] Add methods to make a user join/leave a group

fperot74:
Uses cloudtrust/keycloak-client#53

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  version = "v1.3.0"
+  branch = "ct-2402-assigngroupstousers"
 
 [[constraint]]
   name = "github.com/go-kit/kit"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  branch = "ct-2402-assigngroupstousers"
+  branch = "master"
 
 [[constraint]]
   name = "github.com/go-kit/kit"
--- a/pkg/management/authorization.go
+++ b/pkg/management/authorization.go
@@ -37,6 +37,7 @@ var (
 	MGMTGetUserAccountStatus                = newAction("MGMT_GetUserAccountStatus", security.ScopeGroup)
 	MGMTGetRolesOfUser                      = newAction("MGMT_GetRolesOfUser", security.ScopeGroup)
 	MGMTGetGroupsOfUser                     = newAction("MGMT_GetGroupsOfUser", security.ScopeGroup)
+	MGMTSetGroupsToUser                     = newAction("MGMT_SetGroupsToUser", security.ScopeGroup)
 	MGMTGetAvailableTrustIDGroups           = newAction("MGMT_GetAvailableTrustIDGroups", security.ScopeRealm)
 	MGMTGetTrustIDGroups                    = newAction("MGMT_GetTrustIDGroups", security.ScopeGroup)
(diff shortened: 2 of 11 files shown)


pkg/management/component.go
```diff
@@ -33,6 +34,8 @@ type KeycloakClient interface {
 	DeleteUser(accessToken string, realmName, userID string) error
 	GetUser(accessToken string, realmName, userID string) (kc.UserRepresentation, error)
 	GetGroupsOfUser(accessToken string, realmName, userID string) ([]kc.GroupRepresentation, error)
+	AddGroupOfUser(accessToken string, realmName, userID, groupID string) error
```
bsoniam: same as in the other PR:
AddGroupToUser?
DeleteGroupFromUser?

pkg/management/endpoint.go
```diff
@@ -67,102 +68,51 @@ type Endpoints struct {
 	CreateShadowUser endpoint.Endpoint
 }
 
-// ManagementComponent is the interface of the component to send a query to Keycloak.
-type ManagementComponent interface {
```
bsoniam: Why is this removed?

fperot74: It is highly linked with Component interface from component.go: if you add a method in one of the two interfaces, you have to add it to the other one. They always have to be exactly duplicated. So it is not required to have both interfaces: we can use only one.

pkg/management/component.go
```diff
@@ -508,6 +512,41 @@ func (c *component) GetGroupsOfUser(ctx context.Context, realmName, userID strin
...
+	var alreadyActiveGroups = make(map[string]bool)
+	for _, grp := range groupsKc {
+		if !validation.IsStringInSlice(groupIDs, *grp.Id) {
+			err = c.keycloakClient.DeleteGroupFromUser(accessToken, realmName, userID, *grp.Id)
+			if err != nil {
+				c.logger.Warn(ctx, "err", err.Error())
+				return err
+			}
+		} else {
+			alreadyActiveGroups[*grp.Id] = true
```
harture: Just a note as it has no functional impact at all: I think it is not really needed to have such array. all those elements will already be in groupsKc so you could use groupKc in the next loop instead of this new array right ?

pkg/management/authorization.go
```diff
@@ -253,6 +254,17 @@ func (c *authorizationComponentMW) GetGroupsOfUser(ctx context.Context, realmNam
 	return c.next.GetGroupsOfUser(ctx, realmName, userID)
 }
 
+func (c *authorizationComponentMW) SetGroupsToUser(ctx context.Context, realmName, userID string, groupIDs []string) error {
+	var action = MGMTSetGroupsToUser.String()
+	var targetRealm = realmName
+
+	if err := c.authManager.CheckAuthorizationOnTargetUser(ctx, action, targetRealm, userID); err != nil {
```
harture: I had not anticipated it but we would need a way to filter the list of group which are assignable by someone. I would suggest to add a new action MGMTAssignableGroupsToUser and add another check in this method which would ensure the groupIDs in parameters are among the allowed ones. 
What do you think of it ?

pkg/management/component.go
fperot74: Already discussed together... This is not an array but a map so it will be slight more performant to check in a map rather than in a slice when executing the second loop.

pkg/management/authorization.go
fperot74: Code updated... please check once again

Gopkg.toml
```diff
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  version = "v1.3.0"
+  branch = "master"
```
harture: should we create a version a keycloak-client ?

PR #206, opened 2020-04-01
[CLOUDTRUST-2458] renaming CreateShadowUser to LinkShadowUser

bsoniam:
Remark: this change requires also a change in the authorizations table as the name of the action has changed from MGMT_CreateShadowUser to MGMT_LinkShadowUser.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  branch = "CT-2458"
+  branch = "master"
 
 [[constraint]]
   name = "github.com/go-kit/kit"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  branch = "master"
+  branch = "CT-2458"
 
 [[constraint]]
   name = "github.com/go-kit/kit"
--- a/configs/keycloak_bridge.yml
+++ b/configs/keycloak_bridge.yml
@@ -90,11 +90,11 @@ db-audit-ro-migration: false
 db-audit-ro-migration-version: 0.1
 db-audit-ro-connection-check: false
 
-# DB Configuration RW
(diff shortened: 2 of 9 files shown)


Gopkg.toml
```diff
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  branch = "master"
+  branch = "CT-2458"
```
fperot74, line 34: Don't forget to restore before merging

configs/keycloak_bridge.yml
```diff
@@ -90,11 +90,11 @@ db-audit-ro-migration: false
 db-audit-ro-migration-version: 0.1
 db-audit-ro-connection-check: false
 
-# DB Configuration RW
+# DB Cloudtrust RW
 db-config-rw-host-port: 172.17.0.2:3306
 db-config-rw-username: bridge
 db-config-rw-password: bridge-password
-db-config-rw-database: configuration
+db-config-rw-database: cloudtrust
```
fperot74, line 97: Do we really want to rename this database? (already in production, referenced in multiple projects, ...)

harture: I have also some concerns about compatibility and I don't have the feeling the expresiveness of the content of the DB is improved with this new name

configs/keycloak_bridge.yml
```diff
@@ -104,11 +104,11 @@ db-config-rw-migration: false
 db-config-rw-migration-version: 0.1
 db-config-rw-connection-check: true
 
-# DB Configuration RO
+# DB Cloudtrust RO
```
harture, line 107: Same as before

configs/keycloak_bridge.yml
bsoniam: Hello, 
I had the impression that the DB is actually called cloudtrust and not configuration. That's why I made these changes. 
But if want to keep the "configuration" name, it is fine by me.

Gopkg.toml
bsoniam: Yes, indeed

PR #207, opened 2020-04-01
[CLOUDTRUST-2357] Attack detection status

fperot74:


pkg/management/endpoint.go
```diff
@@ -440,6 +441,15 @@ func MakeClearUserLoginFailures(component Component) cs.Endpoint {
 	}
 }
 
+// MakeGetAttackDetectionStatus creates an endpoint for GetAttackDetectionStatus
+func MakeGetAttackDetectionStatus(component Component) cs.Endpoint {
+	return func(ctx context.Context, req interface{}) (interface{}, error) {
+		var m = req.(map[string]string)
+
```
bsoniam, line 448: credentialID is in the body?
or maybe there is no credentialID

api/management/swagger-api_management.yaml
```diff
@@ -761,6 +761,39 @@ paths:
...
+        required: true
+        schema:
+          type: string
+      - name: userID
+        in: path
+        description: User id
+        required: true
+        schema:
+          type: string
+      - name: credentialID
```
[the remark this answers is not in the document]
fperot74: Right. I removed it (copied/pasted without reason)

pkg/management/endpoint.go
fperot74: credentialID was an error in the swagger file.

PR #213, opened 2020-04-09
[CLOUDTRUST-2506] Avoid panic when no-op database is used

fperot74:
+ Apply lint recommendations

### DIFF
--- a/cmd/keycloakb/keycloak_bridge.go
+++ b/cmd/keycloakb/keycloak_bridge.go
@@ -40,6 +40,8 @@ import (
 	"github.com/cloudtrust/keycloak-bridge/pkg/statistics"
 	"github.com/cloudtrust/keycloak-bridge/pkg/validation"
 	keycloak "github.com/cloudtrust/keycloak-client"
+	keycloakapi "github.com/cloudtrust/keycloak-client/api"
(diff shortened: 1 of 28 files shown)


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -122,17 +122,17 @@ func main() {
...
 		// DB - for the moment used just for audit events
-		auditRwDbParams = database.GetDbConfig(c, "db-audit-rw", !c.GetBool("events-db"))
+		auditRwDbParams = database.GetDbConfig(c, "db-audit-rw")
 
 		// DB - Read only user for audit events
-		auditRoDbParams = database.GetDbConfig(c, "db-audit-ro", false)
+		auditRoDbParams = database.GetDbConfig(c, "db-audit-ro")
 
 		// DB for custom configuration
-		configRwDbParams = database.GetDbConfig(c, "db-config-rw", !c.GetBool("config-db-rw"))
```
harture, line 131: for example, config-db-rw is still set at some point in this file. I imagine we should remove it ? (same for the others)

fperot74: Ok... I just renamed fields when initializing the configuration.

PR #215, opened 2020-04-14
CLOUDTRUST-2378 More IE11 headers

fratt-elca:


fratt-elca:
Reason : https://stackoverflow.com/a/51097426

PR #219, opened 2020-04-20
[CLOUDTRUST-2237] Encrypt user data

bsoniam:
PII stored by Keycloak-Bridge in users DB:
- user_details table (details)
- checks table (proof_data)
are now encrypted

This comes with a change in the DB schema of user_details. 
Small corrections are applied to Open api 3.0.


Gopkg.toml
```diff
@@ -27,11 +27,13 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/common-service"
-  version = "v2.1.1"
+#  version = "v2.1.0"
+  branch = "CT-2237-encrypt-PII"
```
fperot74, line 31: Don't forget to use branch="master" or use a new version release before merging

api/validation/swagger_api_validation.yml
```diff
@@ -30,7 +30,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/User'
-    put:
+    post:
```
fperot74, line 33: The summary tells "Updates user data information"... We should either use put method with verb update in summary or use post method with verb create in summary

api/validation/swagger_api_validation.yml
```diff
@@ -107,12 +107,13 @@ components:
...
         datetime:
-          type: integer
-          format: int64
+          type: string
+          format: date-time
         status:
           type: string
         proofData:
           type: string
+          format: byte
```
fperot74, line 116: byte? Type is string... Did you intend to write something like base64 or something like that?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -235,6 +235,13 @@ func main() {
 		}
 	}
 
+	// Security - AES encryption mechanism for users PII
+	aesEncryption, err := security.NewAesGcmEncrypterFromBase64(c.GetString("db-aesgcm-key"), c.GetInt("db-aesgcm-tag-size"))
```
fperot74, line 239: Please check my PR: https://github.com/cloudtrust/keycloak-bridge/pull/217/files#diff-87b64553cc8dcbe43a0fce63c71a5c71R81
You can declare CfgDbAesGcmKey and CfgDbAesGcmTagSize constants and declare/initialize the configuration in func config(...)

fperot74: Referenced PR is merged...

api/validation/swagger_api_validation.yml
bsoniam: Well, we do a POST so I changed the verb in the summary

api/validation/swagger_api_validation.yml
bsoniam: https://swagger.io/docs/specification/data-models/data-types/  (check String formats)
A base64 string is a string with format byte

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -1378,6 +1387,10 @@ func config(ctx context.Context, logger log.Logger) *viper.Viper {
 	v.BindEnv(CfgValidationBasicAuthToken, "CT_BRIDGE_VALIDATION_BASIC_AUTH")
 	censoredParameters[CfgValidationBasicAuthToken] = true
 
+	v.BindEnv(CfgDbAesGcmKey, "CT_BRIDGE_DB_AES_KEY")
```
harture, line 1390: missing setDefault

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -1378,6 +1387,10 @@ func config(ctx context.Context, logger log.Logger) *viper.Viper {
...
+	censoredParameters[CfgDbAesGcmKey] = true
+	v.SetDefault(CfgDbAesGcmTagSize, 16)
```
harture, line 1392: Would be nice to move setDefault with all the others

internal/keycloakb/usersdbmodule_test.go
```diff
@@ -108,6 +182,24 @@ func TestGetUserInformation(t *testing.T) {
...
 	})
 
+	t.Run("Error at decryption", func(t *testing.T) {
+		var natureValue = "nature"
+		gomock.InOrder(
+			mockDB.EXPECT().Query(gomock.Any(), realm, userID).Return(mockSQLRows, nil),
+			mockSQLRows.EXPECT().Next().Return(true),
+			mockSQLRows.EXPECT().Scan(gomock.Any()).DoAndReturn(func(checkID *int64, realm *string, userID *string, operator *sql.NullString,
+				datetime *sql.NullString, status *sql.NullString, checkType *sql.NullString, nature *sql.NullString, proofType *sql.NullString,
+				proofData *[]byte, comment *sql.NullString) error {
```
fperot74, line 192: You can do it simpler:
* for each unused parameter, you can use _ rather than naming it and user interface{}
Then, you can simplify the writing interface{} parameters:
* `_ interface{}, _ interface{}, _ interface{}` becomes `_, _, _ interface{}`

You finally get something like:
`mockSQLRows.EXPECT().Scan(gomock.Any()).DoAndReturn(func(_, _, _, _, _, _, _ interface{}, nature *sql.NullString, _, _, _ interface{}) error {`

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -312,6 +314,13 @@ func main() {
 		}
 	}
 
+	// Security - AES encryption mechanism for users PII
```
harture, line 317: Would be good to have a check which ensure the encryption key is not empty

configs/keycloak_bridge.yml
```diff
@@ -192,4 +196,4 @@ register-techuser-client-id: admin-cli
 register-enduser-client-id: selfserviceid
 recaptcha-url: https://www.google.com/recaptcha/api/siteverify
 recaptcha-secret: theverymysterioussecretfortherecaptchaverifyoperation
-sse-public-url: https://sse.trustid.ch
+sse-public-url: https://sse.trustid.ch
```
harture, line 199: why do we remove the new line ? Usually it is prefered to have one no ?

pkg/validation/component.go
```diff
@@ -164,6 +164,7 @@ func (c *component) UpdateUser(ctx context.Context, userID string, user api.User
 
 		var userDB = dto.DBUser{
 			UserID:           &uID,
+			BirthLocation:    user.BirthLocation,
```
harture, line 167: Why this is added ?

bsoniam: Because it was missing. The struct contains the birthlocation but the code was missing it. 
If there is a reason we should exclude it, let me know @harture .

cmd/keycloakb/keycloak_bridge.go
fperot74: Isn't there an automatic test of encrypt/decrypt using the key ?

internal/keycloakb/usersdbmodule_test.go
bsoniam: OK

cmd/keycloakb/keycloak_bridge.go
bsoniam: OK

cmd/keycloakb/keycloak_bridge.go
bsoniam: DONE

cmd/keycloakb/keycloak_bridge.go
fperot74: We are performing an encryption/decryption cycle using the key in NewAesGcmEncrypterFromBase64(...)
That's why there is a returned error. I suppose that if the encryption key is empty, the encryption/decryption cycle should fail.

bsoniam: No, actually if the key is "" then the decoding succeeds (I just checked). I am adding a modification on the common-service to ensure that the correct length of the key is provided.

configs/keycloak_bridge.yml
bsoniam: DONE

cmd/keycloakb/keycloak_bridge.go
harture: If we do it at startup it would avoid to fail while performing the operation, that's why it could be interesting to have such check. As you want not an hard requirement

fperot74: The check is already done in NewAesGcmEncrypterFromBase64(...)

pkg/validation/component.go
harture: ok

harture:
Rebased on master 23.06.2020

PR #235, opened 2020-06-29
[CLOUDTRUST-2691] Add barcode type in realm configuration

fperot74:
Requires:
* cloudtrust/common-service#57
* cloudtrust/keycloak-client#64

### DIFF
--- a/api/account/api.go
+++ b/api/account/api.go
@@ -58,6 +58,7 @@ type Configuration struct {
 	ShowAccountDeletionButton         *bool           `json:"show_account_deletion_button"`
 	RedirectSuccessfulRegistrationURL *string         `json:"redirect_successful_registration_url"`
 	AvailableChecks                   map[string]bool `json:"available-checks"`
+	BarcodeType                       *string         `json:"barcode_type"`
 }
 
 // UpdatePasswordBody is the definition of the expected body content of UpdatePassword method
--- a/api/account/api.go
+++ b/api/account/api.go
@@ -112,26 +112,30 @@ func ConvertToAPIAccount(ctx context.Context, userKc kc.UserRepresentation, logg
 		userRep.Locale = value
 	}
 	if values := userKc.GetAttribute(constants.AttrbAccreditations); len(values) > 0 {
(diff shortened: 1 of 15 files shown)


api/account/api.go
```diff
@@ -111,26 +112,30 @@ func ConvertToAPIAccount(ctx context.Context, userKc kc.UserRepresentation, logg
...
+	var accreds []AccreditationRepresentation
+	for _, accredJSON := range values {
+		var accred AccreditationRepresentation
+		if json.Unmarshal([]byte(accredJSON), &accred) == nil {
+			// If accreditation is revoked, considers it as expired so that this version is backward compatible with backoffice
+			if accred.Revoked != nil && *accred.Revoked {
+				accred.Expired = accred.Revoked
 			} else {
-				logger.Warn(ctx, "msg", "Can't unmarshall JSON", "json", accredJSON)
+				accred.Expired = keycloakb.IsDateInThePast(accred.ExpiryDate)
```
harture, line 130: I imagine this a bug for a different topic, it might be good to have it in a specific PR to be able to track it in future ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -1272,33 +1273,21 @@ func config(ctx context.Context, logger log.Logger) *viper.Viper {
 	v.SetDefault(CfgTimeout, "5s")
 
 	// Storage events in DB (read/write)
-	v.SetDefault(CfgAuditRwDbParams+"-enabled", false)
```
harture, line 1275: Why do we remove all those things ?

api/account/api.go
fperot74: No, it's just the diff which is not correctly computed by GitHub... the warning is still present (see L#135)

cmd/keycloakb/keycloak_bridge.go
fperot74: ConfigureDbDefault already initializes the -enabled parameter with value true...
This one is moved 2 lines below to set the parameter to false

PR #236, opened 2020-07-02
[CLOUDTRUST-2669] Allow dot character in ID document number

fperot74:


internal/constants/regexp.go
```diff
@@ -23,6 +23,7 @@ const (
 	RegExpGender           = `^[MF]$`
 	RegExpBirthDate        = `^(\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))$`
 	RegExpLocale           = `^[a-z]{2}$`
+	RegExpIDDocumentNumber = `^([\w\d]+([\. -][\w\d]+)*)$`
```
harture, line 26: Why do we remove {1,50} ?

fperot74: Reverted

PR #240, opened 2020-07-14
Add extra fields in user API

harture:


pkg/management/component.go
```diff
@@ -304,7 +326,12 @@ func (c *component) DeleteUser(ctx context.Context, realmName, userID string) er
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 
 	err := c.keycloakClient.DeleteUser(accessToken, realmName, userID)
+	if err != nil {
```
fperot74, line 329: Would it be better to create a transaction, delete the user in DB, delete the user in KC and rollback/commit if there is/isn't error? (same question for CreateUser)

pkg/management/component.go
```diff
@@ -335,11 +362,26 @@ func (c *component) GetUser(ctx context.Context, realmName, userID string) (api.
 		username = *userKc.Username
 	}
 
+	// Retrieve info from DB user
+	dbUser, err := c.usersDBModule.GetUser(ctx, realmName, *userKc.ID)
+	if err != nil {
```
fperot74, line 367: If user is not found, is err nil? We should not return an error if err==sql.ErrNoRows...
(according to the following if statement, I thing that both dbUser and err are nil in such a case)

pkg/management/component.go
```diff
@@ -294,6 +303,19 @@ func (c *component) CreateUser(ctx context.Context, realmName string, user api.U
 	reg := regexp.MustCompile(`[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}`)
 	userID := string(reg.Find([]byte(locationURL)))
 
+	// Store user in database
+	err = c.usersDBModule.StoreOrUpdateUser(ctx, realmName, dto.DBUser{
+		UserID:               &userID,
+		BirthLocation:        user.BirthLocation,
```
fperot74, line 309: If all fields are nil, do we need to create a row in DB?

pkg/management/component.go
```diff
@@ -361,6 +403,15 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 	}
 	keycloakb.ConvertLegacyAttribute(&oldUserKc)
 
+	// get the "old" user infos from database
+	var oldDbUser *dto.DBUser = &dto.DBUser{}
```
fperot74, line 407: You should set UserID in case user is not found in DB

pkg/management/component.go
harture: That's good point, I will improve it.

pkg/management/component.go
harture: I don't understand the point. If user is not found, the method returns (nil,nil) so if there is an error I return it and few lines after I manage the case if the is no rows (if dbUser == nil).

pkg/management/component.go
harture: there is pro and cons about it .... I thought it would make sense to create it anyway but your comment make me think about it a second time and I think you are right.

pkg/management/component.go
harture: I am asking my self if it should not be the job of the module to return such stuct... Would you be against such modification ?

pkg/management/component.go
harture: I think we need to discuss about transactions in general when we will be both back from holidays. So I will temporary postpone this improvement.

pkg/management/component.go
harture: done

pkg/management/component.go
fperot74: If the user is not found, it would be strange to return a default DBUser value without information that user was not found.

PR #241, opened 2020-07-15
Cloudtrust 2660 sync user representation fields

harture:


api/kyc/api.go
```diff
@@ -19,13 +19,13 @@ type ActionRepresentation struct {
 
 // UserRepresentation contains user details
 type UserRepresentation struct {
-	UserID               *string                        `json:"userId,omitempty"`
+	ID                   *string                        `json:"id,omitempty"`
```
fperot74, line 22: Good point... Is front end updated accordingly?

PR #246, opened 2020-08-20
[CLOUDTRUST-2767] Fix endpoint availability check when validating a user from master

fperot74:


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -831,6 +831,11 @@ func main() {
 		}
 	}
 
+	// Tools for endpoint middleware
+	var idRetriever = keycloakb.NewRealmIDRetriever(keycloakClient)
+	var configurationReaderDBModule = configuration.NewConfigurationReaderDBModule(configurationRoDBConn, logger)
+	var endpointAvailabilityChecker = middleware.NewEndpointAvailabilityChecker(configuration.CheckKeyPhysical, idRetriever, configurationReaderDBModule)
```
harture, line 837: maybe endpointPhysicalCheckAvailabilityChecker ?

harture (approved):

cmd/keycloakb/keycloak_bridge.go
fperot74: Why not... even if I'm not fan of long variable names

PR #248, opened 2020-08-27
[CLOUDTRUST-2777] Users archiving

fperot74:
If this PR is merged after cloudtrust/keycloak-bridge#247, IDDocumentCountry might be added to archived fields

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ CT_BRIDGE_DB_CONFIG_USERNAME | db-config-username
 CT_BRIDGE_DB_CONFIG_PASSWORD | db-config-password
 CT_BRIDGE_DB_USERS_USERNAME | db-users-username
 CT_BRIDGE_DB_USERS_PASSWORD | db-users-password
+CT_BRIDGE_DB_ARCHIVE_RW_USERNAME | db-archive-rw-username
+CT_BRIDGE_DB_ARCHIVE_RW_PASSWORD | db-archive-rw-password
 CT_BRIDGE_INFLUX_USERNAME | influx-username
 CT_BRIDGE_INFLUX_PASSWORD | influx-password
 CT_BRIDGE_SENTRY_DSN | sentry-dsn
--- a/cmd/keycloakb/keycloak_bridge.go
+++ b/cmd/keycloakb/keycloak_bridge.go
@@ -132,6 +132,9 @@ const (
 	cfgSsePublicURL             = "sse-public-url"
 	cfgDbAesGcmKey              = "db-aesgcm-key"
 	cfgDbAesGcmTagSize          = "db-aesgcm-tag-size"
+	cfgArchiveRwDbParams        = "db-archive-rw"
+	cfgDbArchiveAesGcmKey       = "db-archive-aesgcm-key"
+	cfgDbArchiveAesGcmTagSize   = "db-archive-aesgcm-tag-size"
 )
 
 func init() {
@@ -208,6 +211,9 @@ func main() {
 
 		// DB for users
 		usersRwDbParams = database.GetDbConfig(c, cfgUsersRwDbParams)
+
+		// DB for archiving users
+		archiveRwDbParams = database.GetDbConfig(c, cfgArchiveRwDbParams)
 
 		// Rate limiting
 		rateLimit = map[RateKey]int{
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -82,6 +82,8 @@ CT_BRIDGE_DB_CONFIG_USERNAME | db-config-username
 CT_BRIDGE_DB_CONFIG_PASSWORD | db-config-password
 CT_BRIDGE_DB_USERS_USERNAME | db-users-username
 CT_BRIDGE_DB_USERS_PASSWORD | db-users-password
+CT_BRIDGE_DB_ARCHIVE_RO_USERNAME | db-archive-ro-username
```
harture, line 85: As it is for archiving, I assume it is not a read-only (ro) user but rw or wo ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -855,7 +880,7 @@ func main() {
 		}
 
 		// new module for KYC service
-		kycComponent := kyc.NewComponent(technicalTokenProvider, registerRealm, keycloakClient, usersDBModule, eventsDBModule, accredsModule, kycLogger)
+		kycComponent := kyc.NewComponent(technicalTokenProvider, registerRealm, keycloakClient, usersDBModule, archiveDBModule, eventsDBModule, accredsModule, kycLogger)
```
harture, line 883: Users can also be modified by validation API (IDnow), so maybe we should also have archiveDBModule for it no?

fperot74: Done

pkg/validation/component.go
```diff
@@ -135,6 +129,13 @@ func (c *component) GetUser(ctx context.Context, realmName string, userID string
 }
 
 func (c *component) UpdateUser(ctx context.Context, realmName string, userID string, user api.UserRepresentation) error {
+	var validationCtx = validationContext{
+		ctx:       ctx,
```
harture, line 133: I feel quite uncomfortable with such architecture ...
component provided inside another struct doesn't follow our currentway of doing things until now. Do we really need to perform such thing ? Why the methods of ValidationContext are not just private method of this component ?
Maybe I am missing one of the constraint

internal/keycloakb/archivedbmodule.go
```diff
@@ -0,0 +1,49 @@
...
+// NewArchiveDBModule returns an archive DB module.
+func NewArchiveDBModule(db sqltypes.CloudtrustDB, cipher security.EncrypterDecrypter, logger log.Logger) ArchiveDBModule {
+	return &archiveDBModule{
+		db:     db,
+		cipher: cipher,
+		logger: logger,
+	}
+}
+
+func (a *archiveDBModule) StoreUserDetails(ctx context.Context, realm, userID string, details []byte) error {
```
harture, line 38: I think it would make sense we provide a UserRepresentation to this module and it will be module responsibility to convert it into byte array.

pkg/validation/component.go
```diff
@@ -295,3 +306,104 @@ func (c *component) reportEvent(ctx context.Context, apiCall string, values ...s
...
+			v.dbUser = &dbUser
+		} else {
+			v.component.logger.Warn(v.ctx, "msg", "Can't get user from database", "err", err.Error(), "realm", v.realmName, "user", v.userID)
+			return nil, err
+		}
+	}
+	return v.dbUser, nil
+}
+
+func (v *validationContext) getUserWithAccreditations() (*kc.UserRepresentation, error) {
```
harture, line 374: I don't understnad why we have to eprform a lot more stuff in validation compare to kyc. 
Maybe we could move all this logic linked to the user archiving into module layer ?

pkg/validation/component.go
fperot74: I removed component from ValidationContext and replaced func (v *ValidationContext) xxx(...) by func (c *component) xxx(v *ValidationContext, ...)

pkg/validation/component.go
fperot74: KYC only stores a kyc/UserRepresentation : all input fields are mandatory
Validation stores a validation/UserRepresentation merged with information from Keycloak and from database then add checks... input fields are not mandatory. A simple way to do this would be to keep the processings as their are already coded then just go and get information in Keycloak and database but it would be much less performant.
I'll try to do it a different way.

internal/keycloakb/archivedbmodule.go
fperot74: see next answer

pkg/validation/component.go
```diff
@@ -148,7 +146,7 @@ func (c *component) UpdateUser(ctx context.Context, realmName string, userID str
 	}
 
 	if kcUpdate || shouldRevokeAccreditations {
-		err = c.updateUserKeycloak(ctx, realmName, userID, user, shouldRevokeAccreditations)
+		err = c.updateUserKeycloak(validationCtx, user, shouldRevokeAccreditations)
```
harture, line 149: It is strange to provide a validation context while we are updating keycloak user. Even if it is the same fields which are part of it I think it is a bit misleading when we read it.

harture (approved):

PR #250, opened 2020-09-18
[CLOUDTRUST-1034] Add methods to lock/unlock a user in management API

fperot74:


pkg/management/component.go
```diff
@@ -506,6 +493,61 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 	return nil
 }
 
+func (c *component) reportLockEvent(ctx context.Context, realmName, userID string, username *string, enabled bool) {
+	var blank = ""
+	if username == nil {
+		username = &blank
```
harture, line 499: As username is a reference, I think we are changing the value of user.Username which seems a bit risky as it may have side effect.

harture (approved):

pkg/management/component.go
fperot74: There is no change in behavior: username is provided if available otherwise empty string is provided.

PR #251, opened 2020-09-29
[CLOUDTRUST-2805] Support multiple realms for register interface

fperot74:
Requires:
* cloudtrust/common-service#62
* cloudtrust/keycloak-client#70

### DIFF
--- a/api/register/swagger-api_register.yaml
+++ b/api/register/swagger-api_register.yaml
@@ -13,12 +13,42 @@ paths:
     post:
       tags:
       - Register
-      summary: Creates a user
+      summary: Creates a user in the configured realm (register-realm)
       security:
         - BasicAuth: [recaptcha]
       parameters:
       - name: realm
         in: query
+        description: realm name (not id!) of a realm configured with a redirect URL for the end of the process
+        required: true
+        schema:
+          type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/User'
+      responses:
+        200:
+          description: Successful operation. Returns the generated username
+          content:
+            application/json:
+              schema:
+                type: string
+        400:
+          description: Invalid information provided
+        403:
+          description: Invalid RECAPTCHA token
+  /register/realms/{realm}/user:
+    post:
+      tags:
+      - Register
+      summary: Creates a user in the realm specified in URL path
+      security:
+        - BasicAuth: [recaptcha]
+      parameters:
+      - name: realm
+        in: path
         description: realm name (not id!)
         required: true
         schema:
@@ -95,6 +125,9 @@ components:
       properties:
         cancelUrl:
           type: string
+        mode:
+          type: string
+          description: is the register API used for corporate or social mode
   securitySchemes:
     BasicAuth:
       type: http
--- a/configs/keycloak_bridge.yml
+++ b/configs/keycloak_bridge.yml
@@ -184,17 +184,35 @@ livenessprobe-http-timeout: 900
 livenessprobe-cache-duration: 500
 
 # Register API parameters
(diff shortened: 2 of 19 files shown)


api/register/swagger-api_register.yaml
```diff
@@ -39,6 +39,36 @@ paths:
           description: Invalid information provided
         403:
           description: Invalid RECAPTCHA token
+  /register/realms/{realm}/user:
+    post:
+      tags:
+      - Register
+      summary: Creates a user
```
harture, line 46: Creates a user in the realm specified in URL path

api/register/swagger-api_register.yaml
```diff
@@ -39,6 +39,36 @@ paths:
           description: Invalid information provided
         403:
           description: Invalid RECAPTCHA token
+  /register/realms/{realm}/user:
```
harture, line 42: I cannot put my comment on /register/user as nothing as changede but as we now have two modes for registration maybe we should describe in the initial version that user is created in trustID realm and realm parameter is used to determine redirection to use at the end of registration process.

api/register/swagger-api_register.yaml
```diff
@@ -39,6 +39,36 @@ paths:
...
+    post:
+      tags:
+      - Register
+      summary: Creates a user
+      security:
+        - BasicAuth: [recaptcha]
+      parameters:
+      - name: realm
+        in: path
+        description: realm name (not id!)
```
harture, line 52: When we create a user in a specific realm, I don't see a scenario where we would need to use the realm configuration of another realm so I would suggest to remove this parameter. We would always use the realm provided in URL

pkg/register/component.go
```diff
@@ -53,68 +54,96 @@ type ConfigurationDBModule interface {
...
+}
+
+func (r *componentBuilder) AddTargetRealm(realmConf RealmRegisterConfiguration) error {
+	return r.component.addTargetRealm(realmConf)
+}
+
 // Component is the register component interface.
 type Component interface {
-	RegisterUser(ctx context.Context, clientRealmName string, user apiregister.UserRepresentation) (string, error)
+	RegisterUser(ctx context.Context, socialRealmName, clientRealmName string, user apiregister.UserRepresentation) (string, error)
```
harture, line 102: socialRealmName is misleading as in case of register in corporate mode, it won't be socialRealmName. Maybe should we use something like targetRealmName ? targetUserCreationRealmName ?

configs/keycloak_bridge.yml
```diff
@@ -196,6 +196,32 @@ recaptcha-url: https://www.google.com/recaptcha/api/siteverify
 recaptcha-secret: theverymysterioussecretfortherecaptchaverifyoperation
 sse-public-url: https://sse.trustid.ch
 
+# Corporate register
+corp-register-keys:
```
harture, line 200: As we did for validation, I think we should have a technical user in master realm which will be able to create user in any realm. It would avoid such kind of complex configuration. What do you think ?

api/register/swagger-api_register.yaml
fperot74: Here, realm is a path parameter. In first version, realm was a query parameter.

configs/keycloak_bridge.yml
fperot74: So I suggest to remove technical users defined for trustID and new endpoints and use a single one : technical-realm

api/register/swagger-api_register.yaml
harture: ohh yes, sorry for it

configs/keycloak_bridge.yml
harture: Yes, I think it is a good idea

harture (approved):

PR #252, opened 2020-10-05
Fix typo in unknown error message.

vloup:


fperot74:
@fratt-elca, can you check on your side if you are using the value unknowError rather than unknownError?

vloup:
There are references in the backoffice of this error message. Will do a PR there too.

fratt-elca:
Why? this is not even shown to the user...

harture:
What should we do with this PR ?

PR #258, opened 2020-10-22
[CLOUDTRUST-2796] Remove event-receiver

harture:


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -634,9 +544,12 @@ func main() {
 		var eventsLogger = log.With(logger, "svc", "events")
 
 		// module to store API calls of the back office to the DB
-		eventsDBModule := configureEventsDbModule(baseEventsDBModule, influxMetrics, eventsLogger, tracer)
+		eventsRWDBModule := database.NewEventsDBModule(eventsDBConn)
```
fperot74, line 547: As we don't have to provide influx/logger parameters anymore, we can create only one events module and reuse it multiple times... (eventsDBConn acts as a pool of connections to the database)
L570, L657, L722, L765...

fperot74 (approved):

cmd/keycloakb/keycloak_bridge.go
harture: The bridge has been designed to have a kind of silo architecture for each feature (management, account, ....). That's why I think it sounds better to have a specific instance of each module for each silo if possible.
As eventsDBConn acts as a pool, multiple instance of this module would use the same pool so it should have no functional impact while keeping each silo isolated.
What do you think of this approach ?

fperot74: I don't know how is implemented the library, if it can have an impact and if it is possible to have multiple pools with lots of idle connections and one overloaded pool blocking some requests...

PR #261, opened 2020-11-16
[CLOUDTRUST-2856] Send enrollment email API

harture:


api/management/swagger-api_management.yaml
```diff
@@ -735,6 +735,13 @@ paths:
       responses:
         200:
           description: "Code correctly sent by SMS"
+          content:
```
harture, line 738: the new onboarding email API is missing

pkg/register/component.go
```diff
@@ -54,139 +39,137 @@ type ConfigurationDBModule interface {
...
+	// Create new user
+	userID, username, err := c.createUser(ctx, accessToken, targetRealmName, user, *realmConf.SelfRegisterGroupNames, autoLoginToken)
 	if err != nil {
 		return "", err
 	}
 
-	var username, userID string
-	userID, username, err = c.storeUser(ctx, accessToken, targetRealmName, customerRealmName, user, kcUser, realmConf)
+	// Send email
+	var onbaordingRedirectURI = *realmConf.OnboardingRedirectURI
```
fperot74, line 165: typo: onboardingRedirectURI

fperot74 (approved):

pkg/register/component.go
```diff
@@ -197,73 +180,62 @@ func (c *component) RegisterUser(ctx context.Context, targetRealmName, customerR
...
-	if err = c.sendExecuteActionsEmail(ctx, accessToken, targetRealmName, authToken, &kcUser, customerRealmName, userID, realmConf); err != nil {
-		return "", "", err
-	}
-
-	return userID, *kcUser.Username, nil
+	return userID, username, nil
 }
 
-func (c *component) createKeycloakUser(ctx context.Context, accessToken, targetRealmName string, kcUser *kc.UserRepresentation) (string, error) {
+func (c *component) createKeycloakUser(ctx context.Context, accessToken string, realmName string, kcUser kc.UserRepresentation) (string, string, error) {
```
fperot74, line 220: If you keep *kc.UserRepresentation, you can update kcUser with new ID and new username and just return a single error

pkg/management/component.go
```diff
@@ -142,13 +151,14 @@ type component struct {
 	usersDBModule           UsersDetailsDBModule
 	eventDBModule           database.EventsDBModule
 	configDBModule          keycloakb.ConfigurationDBModule
+	onboardingModule        keycloakb.OnboardingModule
```
fperot74, line 154: You don't need to copy/paste OnboardingModule interface in this file if you use keycloakb.OnboardingModule
In the NewComponent, you are using the local OnboardingModule.
You might choose which one you are using for this whole file

api/management/swagger-api_management.yaml
```diff
@@ -713,12 +713,12 @@ paths:
       responses:
         200:
           description: successful operation
-  /realms/{realm}/users/{userID}/send-new-enrolment-code:
+  /realms/{realm}/users/{userID}/send-sms-code:
```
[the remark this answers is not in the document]
harture: Arffff.... I remember I asked if it was ok to avoid to have to rename everything twise ... What do you suggest ?

pkg/management/authorization.go
```diff
@@ -49,7 +49,8 @@ var (
 	MGMTAddClientRolesToUser                = newAction("MGMT_AddClientRolesToUser", security.ScopeGroup)
 	MGMTResetPassword                       = newAction("MGMT_ResetPassword", security.ScopeGroup)
 	MGMTExecuteActionsEmail                 = newAction("MGMT_ExecuteActionsEmail", security.ScopeGroup)
-	MGMTSendNewEnrolmentCode                = newAction("MGMT_SendNewEnrolmentCode", security.ScopeGroup)
+	MGMTSendSmsCode                         = newAction("MGMT_SendSmsCode", security.ScopeGroup)
```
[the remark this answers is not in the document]
harture: I have added a line in the "changes to propagate" so we migrate this right

pkg/register/component.go
harture: tks

pkg/management/component.go
harture: good catch thks

internal/keycloakb/onboardingmodule.go
```diff
@@ -0,0 +1,104 @@
...
+	keycloakClient OnboardingKeycloakClient
+	keycloakURL    string
+	logger         log.Logger
+}
+
+type OnboardingKeycloakClient interface {
+	ExecuteActionsEmail(accessToken string, realmName string, userID string, actions []string, paramKV ...string) error
+}
+
+type OnboardingModule interface {
```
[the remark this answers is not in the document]
harture: In theory this interface is useless and should never be used. However, we have golint issue if we don't have this interface. I think that's why we did it that way everywhere .... If we would have time, it would be interesting to look what is the way to do it as the interface must be outside in Golang

PR #262, opened 2020-11-16
[CLOUDTRUST-2869] Specific validation rules for trustID realm

fperot74:


internal/keycloakb/onboardingmodule.go
```diff
@@ -102,3 +111,50 @@ func (om *onboardingModule) SendOnboardingEmail(ctx context.Context, accessToken
...
+
+	for j := 0; j < length; j++ {
+		nBig, _ := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
+		index := int(nBig.Int64())
+		b.WriteRune(chars[index])
+	}
+	return b.String()
+}
+
+func (om *onboardingModule) CreateUser(ctx context.Context, accessToken, realmName, targetRealmName string, kcUser *kc.UserRepresentation) (string, error) {
```
harture, line 126: I think it is error prone to have side effect on input parameters with methods returning something, especially in our code where we almost never do such thing. 
I would suggest to return (string, string, error) in this specific case. What do you think ?

pkg/management/component.go
```diff
@@ -302,12 +303,20 @@ func (c *component) CreateUser(ctx context.Context, realmName string, user api.U
...
-	userRep = api.ConvertToKCUser(user)
+	var userRep = api.ConvertToKCUser(user)
 
-	// Store user in KC
-	locationURL, err := c.keycloakClient.CreateUser(accessToken, ctxRealm, realmName, userRep)
+	var locationURL string
+	var err error
+	if realmConf, errConf := c.configDBModule.GetAdminConfiguration(ctx, realmName); errConf != nil {
+		err = errConf
+	} else if realmConf.SelfRegisterEnabled != nil && *realmConf.SelfRegisterEnabled {
```
harture, line 312: I don't understand why we have such condition.
We can have selfregister in corporate.
I think the condition should a test on realmName == socialRealm ?

harture (changes requested):

pkg/management/component.go
fperot74: Done

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -584,7 +583,7 @@ func main() {
 
 		var keycloakComponent management.Component
 		{
-			keycloakComponent = management.NewComponent(keycloakClient, usersDBModule, eventsDBModule, configDBModule, onboardingModule, trustIDGroups, managementLogger)
+			keycloakComponent = management.NewComponent(keycloakClient, usersDBModule, eventsDBModule, configDBModule, onboardingModule, trustIDGroups, registerRealm, managementLogger)
```
harture, line 586: a l'occasion est ce que l'on devrait pas renommer register realm en socialRealm or citizenRealm car mainteant on peut faire un register dans n'importe quel realm en fait

harture (approved):

cmd/keycloakb/keycloak_bridge.go
fperot74: Uniquement dans le code ou aussi le nom du paramètre dans la conf ?

PR #263, opened 2020-11-27
[CLOUDTRUST-2893] Add endpoint to check user status

fperot74:


pkg/management/component.go
```diff
@@ -635,6 +636,67 @@ func (c *component) GetUserAccountStatus(ctx context.Context, realmName, userID
...
+
+	var users, err = c.keycloakClient.GetUsers(accessToken, ctxRealm, realmName, "email", email)
+	if err != nil {
+		c.logger.Warn(ctx, "err", "Can't get user by email", "realm", realmName)
+		return kc.UserRepresentation{}, err
+	}
+	// Only search in first page
+	var exactMatch []kc.UserRepresentation
+	if users.Count != nil {
+		for _, user := range users.Users {
```
harture, line 651: Why do we need this ? 
Is it not possible to check len of users.Users and return the first element instead of looping over the whole list ?

harture (approved):

pkg/management/component.go
fperot74: No, it's not... The getUsers() method is a search method and its criterias are intended to be part of the result and not exactly equal to the result.
So if you are requesting white@disney.co, you can get snow.white@disney.com as a result.

harture: I missed this point ... thanks

harture (approved):

PR #266, opened 2020-12-09
[SCRP-32] Communication API

pablo-lo:
This implelents the COM stack for the CPR project. It should be merged at the same time like https://github.com/cloudtrust/keycloak-client/pull/73.

### DIFF
--- a/pkg/communications/endpoint_test.go
+++ b/pkg/communications/endpoint_test.go
@@ -1,0 +1,127 @@
+package communications
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	api "github.com/cloudtrust/keycloak-bridge/api/communications"
+	"github.com/cloudtrust/keycloak-bridge/pkg/communications/mock"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetActionsEndpoint(t *testing.T) {
+	var mockCtrl = gomock.NewController(t)
+	defer mockCtrl.Finish()
+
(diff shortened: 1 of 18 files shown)


harture (commented):
This communications API will be published as Internal. However, it would be great to be able to configure the rights from the BO as for the others. 
Could you add the COM_XXXX into the list of actions returned by the actions endpoint of management with a comment which explain the aim of adding it that way?

pkg/communications/endpoint_test.go
```diff
@@ -0,0 +1,127 @@
...
+	defer mockCtrl.Finish()
+
+	var mockCommunicationsComponent = mock.NewComponent(mockCtrl)
+
+	var e = MakeSendEmailEndpoint(mockCommunicationsComponent)
+
+	var realm = "master"
+	var ctx = context.Background()
+
+	// No error
```
fperot74, line 42: You can use t.Run("test name", func(t *testing.T) { ... })

pkg/communications/http.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/cloudtrust/keycloak-bridge/internal/constants"
+
+	commonhttp "github.com/cloudtrust/common-service/http"
+	"github.com/cloudtrust/common-service/log"
+	"github.com/go-kit/kit/endpoint"
+	http_transport "github.com/go-kit/kit/transport/http"
+)
+
+const (
+	reqScheme = "scheme"
```
fperot74, line 16: Don't need this here (only used in tests)

fperot74 (approved):

harture (approved):

PR #271, opened 2020-12-18
[CLOUDTRUST-2816] Add field in account configuration for supported locales

fperot74:


fratt-elca (changes requested):
- You forgot to add it to /register/config as well
- When internationalizationEnabled = false in the RealmConfiguration returned by Keycloak, you shouldn't return a supportedLocales field

fratt-elca (approved):
Everything works as expected, thank you!
As far as I'm concerned, you can merge when you're ready.

harture (approved):

PR #282, opened 2021-01-26
[CLOUDTRUST-2944] Use target realm conf, not social realm to ensure contact verified

fperot74:


fratt-elca (approved):
Retest ok :
- Corporate scenario
- Social scenario

PR #283, opened 2021-01-27
[CLOUDTRUST-2937] Editing PII must void accreditations (add missing fields)

fperot74:


pkg/management/component.go
```diff
@@ -470,7 +470,16 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 		}
 	}
 
-	var revokeAccreditations = keycloakb.IsUpdated(user.FirstName, oldUserKc.FirstName,
+	// Update in DB user for extra infos
+	// Store user in database
+	var userInfosUpdated = keycloakb.IsUpdated(user.BirthLocation, oldDbUser.BirthLocation,
```
fratt-elca, line 475: birthLocation is not a DEP mandatory field

pkg/management/component.go
```diff
@@ -470,7 +470,16 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
...
+		user.Nationality, oldDbUser.Nationality,
```
fratt-elca, line 476: nationality is not a DEP mandatory field

pkg/management/component.go
```diff
@@ -470,7 +470,16 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
...
+		user.IDDocumentType, oldDbUser.IDDocumentType,
```
fratt-elca, line 477: idDocumentType is not a DEP mandatory field

pkg/management/component.go
```diff
@@ -470,7 +470,16 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
...
+		user.IDDocumentNumber, oldDbUser.IDDocumentNumber,
+		user.IDDocumentExpiration, oldDbUser.IDDocumentExpiration,
```
fratt-elca, line 479: idDocumentExpiration is not a DEP mandatory field

pkg/management/component.go
```diff
@@ -470,7 +470,16 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
...
+		user.IDDocumentCountry, oldDbUser.IDDocumentCountry)
```
fratt-elca, line 480: idDocumentCountry is not a DEP mandatory field

fratt-elca, line 480: idDocumentCountry is not a DEP mandatory field

fratt-elca (changes requested):
Only the DEP fields should void the accreditations :
- firstName
- lastName
- gender
- birthDate
- idDocumentNumber

PS1 : Sorry, the JIRA was not detailed enough, it's my fault.

PS2 : When we will implement the "Attribute management" and "Accreditations management" User Stories, this will be dynamic and each type of accreditations will require their own list of fields, and will be lost if one of those fields is changed. But for now, it's hardcoded for DEP because we're on production with that one.

fratt-elca (approved):
It works.

harture (approved):

PR #286, opened 2021-02-04
[CLOUDTRUST-2964] Use customer theme realm when sending onboarding email

fperot74:


pkg/management/component.go
```diff
@@ -81,7 +81,7 @@ type OnboardingModule interface {
 	GenerateAuthToken() (keycloakb.TrustIDAuthToken, error)
 	OnboardingAlreadyCompleted(kc.UserRepresentation) (bool, error)
 	SendOnboardingEmail(ctx context.Context, accessToken string, realmName string, userID string, username string,
-		autoLoginToken keycloakb.TrustIDAuthToken, onboardingClientID string, onboardingRedirectURI string, reminder bool) error
+		autoLoginToken keycloakb.TrustIDAuthToken, onboardingClientID string, onboardingRedirectURI string, themeRealmName string, reminder bool) error
```
harture, line 84: As there is a new parameter, I would expect a modification to be able to retrieve this new param from the HTTP query.
Do we forgot it or am I missing something ?

harture: Do we have a way to restrict the theme requested to avoid anyone can send an email with the theme of any of our customer ... ?

harture (approved):

Gopkg.toml
```diff
@@ -31,7 +31,8 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  version = "2.3.3"
+#  version = "2.3.3"
+  branch = "ct-2964"
```
harture, line 35: Do not forget it ;)

harture (approved):

PR #289, opened 2021-02-24
[CLOUDTRUST-2997] KYC-Use consent code to protect PII

fperot74:


pkg/kyc/component.go
```diff
@@ -118,33 +123,78 @@ func (c *component) GetUserByUsernameInSocialRealm(ctx context.Context, username
 		return apikyc.UserRepresentation{}, err
 	}
 
+	var realmConf configuration.RealmAdminConfiguration
+	if realmConf, err = c.configDBModule.GetAdminConfiguration(ctx, c.socialRealmName); err != nil {
+		return apikyc.UserRepresentation{}, err
+	}
+	var consentRequired = realmConf.ConsentRequired == nil || *realmConf.ConsentRequired
```
harture, line 130: It might be easier for evereyone (i.e. front) if this methods always return minimum data ?
I imagine the angular app will not optimize call according to the consent config ?

pkg/kyc/component.go
```diff
@@ -118,33 +123,78 @@ func (c *component) GetUserByUsernameInSocialRealm(ctx context.Context, username
...
+			case errorhandler.Error:
+				if e.Status == http.StatusForbidden {
+					return c.createConsentError(errorhandler.MsgErrInvalidQueryParam)
+				}
+			}
+			return err
+		}
+	}
+
+	return err
```
harture, line 173: err should always be nil here. Easier to understand with return nil ?

pkg/kyc/component.go
```diff
@@ -118,33 +123,78 @@ func (c *component) GetUserByUsernameInSocialRealm(ctx context.Context, username
...
+}
+
+func (c *component) checkUserConsent(ctx context.Context, accessToken string, realm string, userID string, consentCode *string) error {
+	var rac, err = c.configDBModule.GetAdminConfiguration(ctx, realm)
+	if err != nil {
+		c.logger.Warn(ctx, "msg", "Can't get realm admin configuration", "realm", realm, "err", err.Error())
+		return err
+	}
+
+	if rac.ConsentRequired == nil || *rac.ConsentRequired {
```
harture, line 157: for retro compatibility, ConsentRequired == nil might means false. What do you think of it ?

pkg/kyc/component.go
```diff
@@ -161,25 +211,43 @@ func (c *component) getUser(ctx context.Context, userID string, kcUser kc.UserRe
...
 
+	if !allDetails {
+		// At this point, we shall not provide too many information if consent is required
+		res = apikyc.UserRepresentation{
+			ID:             res.ID,
+			Username:       res.Username,
+			FirstName:      res.FirstName,
+			LastName:       res.LastName,
+			Email:          res.Email,
+			Accreditations: res.Accreditations,
```
harture, line 222: I would suggest to keep only:
ID, username, firstname, lastname, phone (as you suggested, it might be usefull to display the phone number on which we send it).
Maybe we could just keep only 2 last digits and country indicator of phone number in the result (e.g. +33XXXXXXXX02)
What do you think of it ?

harture (approved):

pkg/kyc/component.go
fperot74: We told it should be activated by default during our daily

pkg/kyc/component.go
fperot74: I added a method to obfuscate phone numbers in common-service

pkg/kyc/authorization.go
```diff
@@ -132,6 +133,17 @@ func (c *authorizationComponentMW) ValidateUser(ctx context.Context, realmName s
 	return c.next.ValidateUser(ctx, realmName, userID, user)
 }
 
+func (c *authorizationComponentMW) SendSmsConsentCodeInSocialRealm(ctx context.Context, userID string) error {
+	var action = KYCSendSmsConsentCodeInSocialRealm.String()
+	var targetRealm = ctx.Value(cs.CtContextRealm).(string)
+
+	if err := c.authManager.CheckAuthorizationOnTargetRealm(ctx, action, targetRealm); err != nil {
```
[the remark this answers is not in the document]
fperot74: Not possible as seen together

pkg/kyc/component.go
fperot74: Done

PR #301, opened 2021-04-08
[CLOUDTRUST-3103] Initialize NameID in mgmt/CreateUser

fperot74:


pkg/management/component.go
```diff
@@ -302,11 +304,20 @@ func (c *component) GetRequiredActions(ctx context.Context, realmName string) ([
...
+func (c *component) CreateUser(ctx context.Context, realmName string, user api.UserRepresentation, generateUsername bool, generateNameID bool) (string, error) {
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 	var ctxRealm = ctx.Value(cs.CtContextRealm).(string)
 
 	var userRep = api.ConvertToKCUser(user)
+	if generateNameID {
+		if oneUUID, errUUID := uuid.NewUUID(); errUUID!=nil {
+			c.logger.Warn(ctx, "msg", "Can't generate UUID", "err", errUUID.Error())
+			return "", errUUID
+		} else {
```
harture, line 316: no need to put it in else {}

harture (approved):

pkg/management/component.go
fperot74: With this syntax ( := in a if statement), affected variables have a scope limited to the if itself. If you don't write it in a else block, you're out of scope and can't use oneUUID anymore.

fperot74: golint also doesn't like it... 2 voices for you, one for me... I'll make a change! ;-)

PR #302, opened 2021-04-12
[CLOUDTRUST-2169] Prevent removal of last MFA credential

fperot74:
+ golint fixes

### DIFF
--- a/internal/keycloakb/mfa_tools.go
+++ b/internal/keycloakb/mfa_tools.go
@@ -1,0 +1,42 @@
+package keycloakb
+
+import (
+	"context"
+
+	errorhandler "github.com/cloudtrust/common-service/errors"
+	kc "github.com/cloudtrust/keycloak-client"
+)
+
+var notMFATypes = map[string]bool{"password": true, "password-history": true}
+
+// CheckRemovableMFA checks if a given credential is removable (owned by user and not the last MFA)
+func CheckRemovableMFA(ctx context.Context, credentialID string, getCredentials func() ([]kc.CredentialRepresentation, error), logger Logger) error {
+	credentialsKc, err := getCredentials()
+	if err != nil {
+		logger.Warn(ctx, "err", err.Error())
+		return err
+	}
+	var found bool
+	var foundRemainingMFA bool
+	for _, credential := range credentialsKc {
+		if credential.ID != nil && *credential.ID != credentialID {
+			if value, ok := notMFATypes[*credential.Type]; !ok || !value {
+				foundRemainingMFA = true
+			}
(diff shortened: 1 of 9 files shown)


internal/keycloakb/kcauthclient_test.go
```diff
@@ -52,8 +52,8 @@ func TestGetGroupNamesOfUserSuccess(t *testing.T) {
 	testKeycloakAuthClient(t, func(t *testing.T, mockKeycloak *mock.KeycloakClient, authClient security.KeycloakClient) {
 		var groupname = "name of group"
 		var groups = []kc.GroupRepresentation{
-			kc.GroupRepresentation{Name: nil},
-			kc.GroupRepresentation{Name: &groupname},
+			{Name: nil},
```
harture, line 55: Cool ??!!! je ne connaissais pas cette syntax ! C'est nouveau ?

internal/keycloakb/mfa_tools.go
```diff
@@ -0,0 +1,41 @@
...
+	credentialsKc, err := getCredentials()
+	if err != nil {
+		logger.Warn(ctx, "err", err.Error())
+		return err
+	}
+	var found bool
+	var foundRemainingMFA bool
+	for _, credential := range credentialsKc {
+		if credential.ID != nil && *credential.ID != credentialID {
+			if value, ok := notMFATypes[*credential.Type]; !ok || !value {
```
harture, line 22: why do we need !ok || ! value ? !ok would not be enough ?

pkg/management/component.go
```diff
@@ -1119,40 +1114,22 @@ func (c *component) GetCredentialsForUser(ctx context.Context, realmName string,
...
+	// Check that credential is removable
+	var err = keycloakb.CheckRemovableMFA(ctx, credentialID, func() ([]kc.CredentialRepresentation, error) {
+		return c.keycloakClient.GetCredentials(accessToken, realmName, userID)
+	}, c.logger)
 	if err != nil {
-		c.logger.Warn(ctx, "msg", "Could not obtain list of credentials", "err", err.Error())
 		return err
 	}
 
-	// Ensure the credential is owned by user
```
harture, line 1129: Ensure the credential is owned by user should remains.

internal/keycloakb/mfa_tools.go
fperot74: In case someone would write `map[string]bool{"password": true, "password-history": true, "ctotp": false}`

pkg/management/component.go
fperot74: What is the use case?

harture (approved):

PR #318, opened 2021-05-25
[CLOUDTRUST-3244] Add lifespan parameter to mgmt/send-onboarding-email

fperot74:


configs/keycloak_bridge.yml
```diff
@@ -210,6 +210,9 @@ technical-username: technical
 technical-password: technicalsuperpasswordverylongandstrong
 technical-client-id: admin-cli
 
+# Maximum lifespan allowed when sending email
+maxLifeSpan: 168h
```
harture, line 214: max-lifespan according to the string defined in keycloak_bridge.go

pkg/management/component.go
```diff
@@ -1029,9 +1031,14 @@ func (c *component) SendSmsCode(ctx context.Context, realmName string, userID st
 	return *smsCodeKc.Code, err
 }
 
-func (c *component) SendOnboardingEmail(ctx context.Context, realmName string, userID string, customerRealm string, reminder bool) error {
+func (c *component) SendOnboardingEmail(ctx context.Context, realmName string, userID string, customerRealm string, reminder bool, lifespan *int) error {
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 
+	if lifespan != nil && *lifespan > c.maxLifeSpan {
+		c.logger.Info(ctx, "msg", "Requested lifespan is too high", "submitted", *lifespan, "considered", c.maxLifeSpan)
```
harture, line 1038: I think we should return a 400 Bad request instead of overriding silently

pkg/management/endpoint.go
```diff
@@ -456,7 +457,16 @@ func MakeSendOnboardingEmailEndpoint(component Component) cs.Endpoint {
 			customerRealmName = value
 		}
 
-		return nil, component.SendOnboardingEmail(ctx, m[prmRealm], m[prmUserID], customerRealmName, reminder)
+		var lifespan *int
+		if value, ok := m[prmQryLifespan]; ok {
+			if iValue, err := strconv.Atoi(value); err == nil {
```
harture, line 462: Could we do the validation that the provided value do not exceed maximum here ?

harture (approved):

PR #333, opened 2021-06-22
[CLOUDTRUST-3279] Add realm display name to the mobile/userinfo endpoint

fperot74:


api/mobile/api.go
```diff
@@ -12,6 +12,7 @@ import (
 
 // UserInformationRepresentation struct
 type UserInformationRepresentation struct {
+	RealmName      *string                        `json:"realmName,omitempty"`
```
harture, line 15: Should we call it DisplayName to avoid confusion later ?

harture (approved):

api/mobile/api.go
fperot74: Renamed: RealmDisplayName

PR #344, opened 2021-09-02
[CLOUDTRUST-3403] Add functions to manage user roles

fperot74:
Needs cloudtrust/keycloak-client#90

### DIFF
--- a/pkg/management/component.go
+++ b/pkg/management/component.go
@@ -43,6 +43,8 @@ type KeycloakClient interface {
 	GetClientRoleMappings(accessToken string, realmName, userID, clientID string) ([]kc.RoleRepresentation, error)
 	AddClientRolesToUserRoleMapping(accessToken string, realmName, userID, clientID string, roles []kc.RoleRepresentation) error
 	GetRealmLevelRoleMappings(accessToken string, realmName, userID string) ([]kc.RoleRepresentation, error)
+	AddRealmLevelRoleMappings(accessToken string, realmName, userID string, roles []kc.RoleRepresentation) error
+	DeleteRealmLevelRoleMappings(accessToken string, realmName, userID string, roles []kc.RoleRepresentation) error
 	ResetPassword(accessToken string, realmName string, userID string, cred kc.CredentialRepresentation) error
 	ExecuteActionsEmail(accessToken string, reqRealmName string, targetRealmName string, userID string, actions []string, paramKV ...string) error
 	SendSmsCode(accessToken string, realmName string, userID string) (kc.SmsCodeRepresentation, error)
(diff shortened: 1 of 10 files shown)


pkg/management/component.go
```diff
@@ -789,6 +793,76 @@ func (c *component) GetRolesOfUser(ctx context.Context, realmName, userID string
...
+	}
+	var roles []kc.RoleRepresentation
+	if role, ok := rolesMap[roleID]; ok {
+		roles = append(roles, role)
+	}
+	if len(roles) == 0 {
+		c.logger.Info(ctx, "msg", "Unknown role", "realm", realmName, "user", userID, "role", roleID)
+		return errorhandler.CreateBadRequestError("role")
+	}
+	err = c.keycloakClient.AddRealmLevelRoleMappings(accessToken, realmName, userID, roles)
```
harture, line 838: I think it is not needed to also send the already mapped roles, just the new one would be needed. For the audit, it would help us to understand was has changed

pkg/management/component.go
```diff
@@ -789,6 +793,76 @@ func (c *component) GetRolesOfUser(ctx context.Context, realmName, userID string
...
+	if err != nil {
+		c.logger.Info(ctx, "msg", "Failed to add user roles", "user", userID, "role", roleID)
+	}
+	return err
+}
+
+func (c *component) DeleteRoleForUser(ctx context.Context, realmName, userID, roleID string) error {
```
harture, line 845: I think the 2 previous comments also applies here.

pkg/management/component.go
```diff
@@ -789,6 +793,76 @@ func (c *component) GetRolesOfUser(ctx context.Context, realmName, userID string
...
+	}
+	var roles []kc.RoleRepresentation
+	if role, ok := rolesMap[roleID]; ok {
+		roles = append(roles, role)
+	}
+	if len(roles) == 0 {
+		c.logger.Info(ctx, "msg", "Unknown role", "realm", realmName, "user", userID, "role", roleID)
+		return errorhandler.CreateBadRequestError("role")
+	}
+	err = c.keycloakClient.DeleteRealmLevelRoleMappings(accessToken, realmName, userID, roles)
```
harture, line 859: I think here we are removing all the roles no ?

pkg/management/component.go
```diff
@@ -789,6 +793,76 @@ func (c *component) GetRolesOfUser(ctx context.Context, realmName, userID string
...
+	var roles, err = c.keycloakClient.GetRealmLevelRoleMappings(accessToken, realmName, userID)
+	if err != nil {
+		c.logger.Info(ctx, "msg", "Failed to get user roles", "realm", realmName, "user")
+		return nil, err
+	}
+	return c.rolesToMap(roles), nil
+}
+
+func (c *component) AddRoleToUser(ctx context.Context, realmName, userID, roleID string) error {
+	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
```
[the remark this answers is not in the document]
fperot74: getRolesAsMap gets existing roles. We get the target one using the provided roleID and add it to a slice. If slice is empty, we return a BadRequestError

pkg/management/component.go
fperot74: We are not sending the already mapped roles... we are sending a slice of roles only containing the role (full struct) whose ID is the provided roleID

pkg/management/component.go
fperot74: I think my 2 previous answers also applies here. ;-)

pkg/management/component.go
fperot74: Only the role whose ID is the provided roleID

harture (approved):

PR #346, opened 2021-09-10
[CLOUDTRUST-3743] Fix default admin configuration with default map for Available checks

basbeu:
Add an default map for available checks
Adapt 1 test in in api_test.go to take the change into account

### DIFF
--- a/api/management/api.go
+++ b/api/management/api.go
@@ -611,6 +611,8 @@ func ConvertRealmAdminConfigurationFromDBStruct(conf configuration.RealmAdminCon
 	var checks = conf.AvailableChecks
 	if checks == nil {
 		checks = make(map[string]bool)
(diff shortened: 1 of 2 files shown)


api/management/api.go
```diff
@@ -611,6 +611,8 @@ func ConvertRealmAdminConfigurationFromDBStruct(conf configuration.RealmAdminCon
 	var checks = conf.AvailableChecks
 	if checks == nil {
 		checks = make(map[string]bool)
+		checks[configuration.CheckKeyIDNow] = false
```
fperot74, line 614: You also can write it with a single instruction:
checks = map[string]bool{configuration.CheckKeyIDNow: false, configuration.CheckKeyPhysical: false}

fperot74 (approved):

harture (approved):

PR #352, opened 2021-10-19
[CLOUDTRUST-3749] Remove the rule which prevent an admin to delete the last authenticator

basbeu:


internal/keycloakb/mfa_tools.go
```diff
@@ -9,8 +9,8 @@ import (
 
 var notMFATypes = map[string]bool{"password": true, "password-history": true}
 
-// CheckRemovableMFA checks if a given credential is removable (owned by user and not the last MFA)
-func CheckRemovableMFA(ctx context.Context, credentialID string, getCredentials func() ([]kc.CredentialRepresentation, error), logger Logger) error {
+// CheckRemovableMFA checks if a given credential is removable (owned by user and not the password credential)
```
harture, line 12: The comment should be adapted as we do not check if the credential is owned by user

harture (approved):

PR #357, opened 2021-11-08
[CLOUDTRUST-3860] Report events ABORTED and CANCELED

basbeu:


api/validation/api.go
```diff
@@ -73,6 +73,9 @@ const (
 	regExpNature      = regExpAlphaNum255
 	regExpProofType   = regExpAlphaNum255
 	regExpStatus      = regExpAlphaNum255
+
+	canceledStatus = "CANCELED"
```
fperot74, line 77: Only used once... do you really need it?

api/validation/api.go
```diff
@@ -73,6 +73,9 @@ const (
...
+	abortedStatus  = "ABORTED"
```
fperot74, line 78: Same here

fperot74 (approved):

PR #359, opened 2021-11-11
[CLOUDTRUST-3873] Forbid uppercase in username

basbeu:


fperot74 (approved):

harture (approved):
Should not be merged for now. We need to ensure that is now ok for all of our clients

PR #361, opened 2021-11-15
[CLOUDTRUST-3768] Denied terms of use management

fperot74:


Gopkg.toml
```diff
@@ -31,7 +31,8 @@
 
 [[constraint]]
   name = "github.com/cloudtrust/keycloak-client"
-  version = "2.4.3"
+  branch = "ct-3768"
+#version = "2.4.3"
```
fperot74, line 35: To be updated before merging

api/tasks/swagger-api_tasks.yaml
```diff
@@ -0,0 +1,23 @@
...
+- url: http://localhost:8888
+tags:
+- name: Tasks
+  description: Tasks management
+paths:
+  /tasks/denied-terms-of-use-users:
+    delete:
+      tags:
+      - Tpackage pkgtasks
+asks
```
basbeu, line 16: I think this word should not be here ! It brokes the swagger syntax

basbeu (approved):

api/tasks/swagger-api_tasks.yaml
fperot74: Good catch ! :-)
That's what happens when you have multiple displays, you type CTRL+C but you are not watching the correct display...

pkg/management/component.go
```diff
@@ -316,13 +316,22 @@ func (c *component) GetRequiredActions(ctx context.Context, realmName string) ([
...
+func (c *component) CreateUser(ctx context.Context, realmName string, user api.UserRepresentation, generateUsername bool, generateNameID bool, termsOfUse bool) (string, error) {
 	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
 	var ctxRealm = ctx.Value(cs.CtContextRealm).(string)
 
 	var userRep = api.ConvertToKCUser(user)
 	userRep.SetAttributeString(constants.AttrbSource, "api")
 
+	if termsOfUse {
+		var reqActions []string
+		if userRep.RequiredActions != nil {
```
harture, line 328: I don't see how it could happen

pkg/tasks/component.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/cloudtrust/common-service/log"
+	api "github.com/cloudtrust/keycloak-bridge/api/tasks"
+	"github.com/cloudtrust/keycloak-bridge/internal/keycloakb"
+	"github.com/cloudtrust/keycloak-client"
+)
+
+// Component interface exposes methods used by the bridge API
+type Component interface {
+	GetActions(ctx context.Context) ([]api.ActionRepresentation, error)
+	DeleteDeniedTermsOfUseUsers(ctx context.Context) error
```
harture, line 18: It is still the old naming

pkg/tasks/component.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (c *component) reportEvent(ctx context.Context, apiCall string, values ...string) {
+	errEvent := c.eventDBModule.ReportEvent(ctx, apiCall, "scheduled-tasks", values...)
+	if errEvent != nil {
+		//store in the logs also the event that failed to be stored in the DB
+		keycloakb.LogUnrecordedEvent(ctx, c.logger, apiCall, errEvent.Error(), values...)
+	}
+}
+
+func (c *component) GetActions(ctx context.Context) ([]api.ActionRepresentation, error) {
```
harture, line 59: Do we have sync with Fratt so we can really set those rights ?

pkg/tasks/component.go
```diff
@@ -0,0 +1,106 @@
...
+		return err
+	}
+
+	// Best effort: in case of error go on trying to remove other users
+	var finalError error
+
+	for _, user := range users {
+		err = c.usersDBModule.DeleteUserDetails(ctx, user.RealmName, user.UserID)
+		if err != nil && err != sql.ErrNoRows {
+			c.logger.Warn(ctx, "msg", "Could not delete database user", "realm", user.RealmName, "usr", user.UserID)
```
harture, line 90: We need to log the error else we won't be able to debug and fix the issue

pkg/tasks/component.go
```diff
@@ -0,0 +1,106 @@
...
+			finalError = err
+		} else {
+			err = c.keycloakClient.DeleteUser(accessToken, user.RealmName, user.UserID)
+			if err != nil {
+				c.logger.Warn(ctx, "msg", "Could not delete keycloak user", "realm", user.RealmName, "usr", user.UserID)
```
harture, line 95: Would be good to log the error here too

pkg/tasks/component.go
```diff
@@ -0,0 +1,106 @@
...
+				finalError = err
+			} else {
+				c.logger.Info(ctx, "msg", "Removed user without terms and conditions acceptance", "realm", user.RealmName, "usr", user.UserID)
+				//store the API call into the DB
+				c.reportEvent(ctx, "API_ACCOUNT_DELETION", database.CtEventRealmName, user.RealmName, database.CtEventUserID, user.UserID)
```
harture, line 100: Is it a newkind of event ? If yes, I have to add it to the list in EPR documentation

pkg/tasks/authorization.go
```diff
@@ -0,0 +1,76 @@
...
+	a := security.Action{
+		Name:  as,
+		Scope: scope,
+	}
+
+	actions = append(actions, a)
+	return a
+}
+
+// Actions used for authorization module
```
harture, line 24: Do we need to put authorization layer with rights for this API ?

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -979,6 +1005,13 @@ func main() {
 		communicationsSubroute.Path("/realms/{realm}/send-mail").Methods("POST").Handler(sendMailHandler)
```
harture, line 1005: it seems we have an indentation issue

harture (approved):

pkg/tasks/component.go
fperot74: No... it's already used in management/deleteUser

pkg/management/component.go
fperot74: Currently, it's not possible... but if someone sets the RequiredActions field with one or more required actions in api.ConvertToKCUser(...), he might not see that his required actions might be ignored as in most cases, he won't set termsOfUse=true

pkg/tasks/component.go
fperot74: We did not speak about a new name here... GetExpiredTermsOfUseAcceptance would no make sense here.
CleanUpAccordingToExpiredTermsOfUseAcceptance ?

pkg/tasks/component.go
fperot74: I removed GetActions in both tasks and communications...
I also added the only action from tasks in the GetActions from management, as we did for actions from communications

cmd/keycloakb/keycloak_bridge.go
fperot74: ???

harture: Interesting... it seems it was a bug in the diff viewer. It is now ok

pkg/tasks/component.go
harture: DeleteUsersWithExpiredTermsOfUseAcceptance ?

pkg/management/component.go
harture: ok

pkg/tasks/authorization.go
fperot74: Discussed together by phone...

PR #374, opened 2022-01-07
[CLOUDTRUST-3966] Adapt bridge for vouchers

basbeu:


internal/keycloakb/accountingclient.go
```diff
@@ -0,0 +1,47 @@
...
+type AccountingClient struct {
+	httpClient HttpClient
+}
+
+type HttpClient interface {
+	Get(data interface{}, plugins ...plugin.Plugin) error
+}
+
+type AccountingBalance struct {
+	Balance *float32 `json:"balance"`
```
fperot74, line 30: float64 is used in accounting (but I don't think it can cause an issue with our values)

internal/keycloakb/accountingclient.go
```diff
@@ -0,0 +1,47 @@
...
+func MakeAccountingClient(httpClient HttpClient) *AccountingClient {
+	return &AccountingClient{
+		httpClient: httpClient,
+	}
+}
+
+func (c *AccountingClient) GetBalance(ctx context.Context, realmName string, userID string, service string) (float32, error) {
+	var correlationID = ctx.Value(cs.CtContextCorrelationID).(string)
+	var res AccountingBalance
+	err := c.httpClient.Get(&res, url.Path(getBalancePath), url.Param(prmRealmName, realmName), url.Param(prmUserID, userID), url.Param(prmService, service), headers.Set(hdrCorrID, correlationID))
```
fperot74, line 42: Could you test it?
Query parameter should be added with query.Add("service", service)... (and it is not part of the URL)

pkg/mobile/component.go
```diff
@@ -146,6 +152,16 @@ func (c *component) GetUserInformation(ctx context.Context) (api.UserInformation
 			c.logger.Debug(ctx, "msg", "User is not allowed to access video identification", "id", ctx.Value(cs.CtContextUserID))
 			delete(availableChecks, actionIDNow)
 		}
+
+		// if vouchers are required to proceed to a video identification, check that the balance is bigger than 1
+		if realmAdminConfig.VideoIdentificationAccountingEnabled != nil && realmAdminConfig.VideoIdentificationPrepaymentRequired != nil &&
```
fperot74, line 157: No need to enter this `if` statement if `actionIDNow` is not in `availableChecks` (avoid 1 HTTP query and associated SQL query behind it)
You also can write a function to be able to write `realmAdminConfig.IsIDNowActivated()` and hide this long checks

fperot74 (approved):

internal/keycloakb/accountingclient.go
basbeu: I will change to have the same type here !

internal/keycloakb/accountingclient.go
basbeu: I tested it ! It works but well I think it is better to do it with  query.Add("service", service) to make it clean.

pkg/mobile/component.go
basbeu: Good catch !

internal/keycloakb/accountingclient.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+
+	cs "github.com/cloudtrust/common-service/v2"
+	"gopkg.in/h2non/gentleman.v2/plugin"
+	"gopkg.in/h2non/gentleman.v2/plugins/headers"
+	"gopkg.in/h2non/gentleman.v2/plugins/url"
+)
+
+const (
+	getBalancePath = "/realms/:realm/users/:userID/balance?service=:service"
```
fperot74, line 13: Missing /internal at the beggining of the path

internal/keycloakb/accountingclient_test.go
```diff
@@ -0,0 +1,50 @@
...
+		assert.Nil(t, err)
+		assert.Equal(t, expectedBalance, balance)
+	})
+
+	t.Run("FAILURE", func(t *testing.T) {
+		mockHttpClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(expectedError).Times(1)
+
+		balance, err := accountingClient.GetBalance(ctx, realmName, userID, service)
+		assert.NotNil(t, err)
+		assert.Equal(t, float32(0), balance)
```
fperot74, line 48: You don't need to test returned value when err is not nil

api/account/api.go
```diff
@@ -77,17 +77,20 @@ type CredentialRepresentation struct {
...
+	ShowPasswordTab                       *bool           `json:"show_password_tab"`
+	ShowProfileTab                        *bool           `json:"show_profile_tab"`
+	ShowAccountDeletionButton             *bool           `json:"show_account_deletion_button"`
+	RedirectSuccessfulRegistrationURL     *string         `json:"redirect_successful_registration_url"`
+	AvailableChecks                       map[string]bool `json:"available-checks"`
+	BarcodeType                           *string         `json:"barcode_type"`
+	Theme                                 *string         `json:"theme"`
+	SupportedLocales                      *[]string       `json:"supportedLocales,omitempty"`
+	ShowGlnEditing                        *bool           `json:"show_gln_editing,omitempty"`
+	VoucherEnabled                        *bool           `json:"voucher_enabled"`
```
harture, line 91: Do we have to expose in account API those 3 parameters ? voucher-enabled is not enough ?

pkg/mobile/component.go
```diff
@@ -146,6 +152,16 @@ func (c *component) GetUserInformation(ctx context.Context) (api.UserInformation
 			c.logger.Debug(ctx, "msg", "User is not allowed to access video identification", "id", ctx.Value(cs.CtContextUserID))
 			delete(availableChecks, actionIDNow)
 		}
+
+		// if vouchers are required to proceed to a video identification, check that the balance is bigger than 1
+		if _, ok := availableChecks[actionIDNow]; ok && realmAdminConfig.VideoIdentificationAccountingEnabled != nil && realmAdminConfig.VideoIdentificationPrepaymentRequired != nil &&
+			*realmAdminConfig.VideoIdentificationAccountingEnabled && *realmAdminConfig.VideoIdentificationPrepaymentRequired {
+			balance, err := c.accountingClient.GetBalance(ctx, realm, userID, "VIDEO_IDENTIFICATION")
```
harture, line 159: log the error

harture (approved):

PR #375, opened 2022-01-12
[CLOUDTRUST-4009] Allow comma in first name values. Make it be configurable

fperot74:


internal/constants/regexp.go
```diff
@@ -53,6 +48,30 @@ const (
...
+	RegExpLastName         = RegExpNameSpecialChars128
+	RegExpCountryCode      = `^\w{2}$`
+	RegExpBirthLocation    = `^([\p{Lu}\p{Ll}][\p{Lu}\p{Ll}\d ()/\\\.\*,'-]{0,49})$`
+	RegExpIDDocumentNumber = `^([\w\d]+([\. -][\w\d]+)*){1,50}$`
+
 	// AllowedDocumentTypes are the valid document type for identification
 	AllowedDocumentTypes = map[string]bool{"ID_CARD": true, "PASSPORT": true, "RESIDENCE_PERMIT": true}
 )
+
+func Init(validationRules map[string]string) {
```
basbeu, line 62: Not a blocking point for me, but maybe Init is not the best name for a function in go, as there is the init() function mechanism.

basbeu (approved):

internal/constants/regexp.go
fperot74: 👍🏻

harture (approved):

PR #378, opened 2022-01-25
[CLOUDTRUST-4014] Handle error timezone

basbeu:


harture (approved):

fperot74 (approved):

pkg/register/component.go
```diff
@@ -203,10 +203,17 @@ func (c *component) sendAlreadyExistsEmail(ctx context.Context, accessToken stri
 	}
 
 	// Add creation date
-	var switzerlandLocation, _ = time.LoadLocation("Europe/Zurich")
-	var creation = time.Unix(creationTimestamp/1000, 0).In(switzerlandLocation)
-	params["creationDate"] = creation.Format("02.01.2006")
-	params["creationHour"] = creation.Format("15:04:05")
+	var creation = time.Unix(creationTimestamp/1000, 0)
+	switzerlandLocation, err := time.LoadLocation("Europe/Zurich")
```
fperot74, line 207: I think it could be better to have a function dedicated for this processing to avoid having it in the middle of the user creation method

basbeu: It is in the sendAlreadyExistsEmail email not in the user creation method !

PR #381, opened 2022-01-28
[CLOUDTRUST-4042] Add creation timestamp in accreditations

basbeu:


fperot74 (approved):

harture (commented):
Have you checked if it is still ok when we return a user ?
In management/component.go, for example we return a user. We load its Keycloak representation and use a converter coming from api/management/api.go, where we manipulate the attributes to determine if a user has some accreditations or not. Not 100% sure if we do not break something there ...

harture (approved):

PR #382, opened 2022-02-01
[CLOUDTRUST-4020] Add a method to revoke a user accreditations

fperot74:


basbeu (approved):

cmd/keycloakb/keycloak_bridge.go
```diff
@@ -1243,6 +1245,7 @@ func main() {
 
 		managementSubroute.Path("/realms/{realm}/users/{userID}/reset-password").Methods("PUT").Handler(resetPasswordHandler)
 		managementSubroute.Path("/realms/{realm}/users/{userID}/execute-actions-email").Methods("PUT").Handler(executeActionsEmailHandler)
+		route.Path("/accreditations/realms/{realm}/users/{userID}/revoke-accreditations").Methods("PUT").Handler(revokeAccreditationsHandler)
```
harture, line 1248: Maybe it would be less hidden that we are in /accreditations if we move it outside of this groups of methods ?

harture (approved):

PR #384, opened 2022-02-03
[CLOUDTRUST-3931] Allow definition of SSE default tab

fperot74:
Dependency not yet added for changes in common-services


api/management/api.go
```diff
@@ -222,6 +223,8 @@ type UserStatus struct {
 const (
 	BOConfKeyCustomers = "customers"
 	BOConfKeyTeams     = "teams"
+
+	regExpSseTabName = `^[a-z]+(-[a-z]+)*$`
```
harture, line 227: As discussed add a amximum size

harture (approved):

api/management/api.go
fperot74: Done

basbeu (approved):

PR #385, opened 2022-02-03
[CLOUDTRUST-3995] Add management API method to update a role

fperot74:
Need to update keycloak-client dependency


harture (approved):

pkg/management/component.go
```diff
@@ -1592,17 +1594,51 @@ func (c *component) CreateRole(ctx context.Context, realmName string, role api.R
 	return locationURL, nil
 }
 
-func (c *component) DeleteRole(ctx context.Context, realmName string, roleID string) error {
-	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
-
+func (c *component) getNonBusinessRole(ctx context.Context, accessToken string, realmName string, roleID string) (kc.RoleRepresentation, error) {
```
basbeu, line 1597: I think this method is rather getBusinessRole().

basbeu: And this method can also used in GetRole method I think !

basbeu (approved):
Do not forget to update the keycloak-client dependency before merging

PR #402, opened 2022-06-03
[CLOUDTRUST-4379] Add delete client role endpoint

basbeu:


pkg/management/component.go
```diff
@@ -2152,6 +2153,24 @@ func (c *component) CreateClientRole(ctx context.Context, realmName, clientID st
 	return locationURL, nil
 }
 
+func (c *component) DeleteClientRole(ctx context.Context, realmName, clientID string, roleID string) error {
+	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
+
+	role, err := c.keycloakClient.GetRole(accessToken, realmName, roleID)
+	if err != nil {
+		c.logger.Warn(ctx, "err", err.Error())
+		return errorhandler.CreateNotFoundError("Role")
```
fperot74, line 2162: I think we always use lower case (or camel case) words in error messages

pkg/management/component.go
```diff
@@ -2152,6 +2153,24 @@ func (c *component) CreateClientRole(ctx context.Context, realmName, clientID st
...
+	}
+
+	if role.ClientRole == nil || role.ContainerID == nil || !*role.ClientRole || *role.ContainerID != clientID {
+		err := errorhandler.CreateNotFoundError("Role")
```
fperot74, line 2166: Same here

pkg/management/component.go
```diff
@@ -2152,6 +2153,24 @@ func (c *component) CreateClientRole(ctx context.Context, realmName, clientID st
...
+		c.logger.Warn(ctx, err, err.Error())
+		return err
+	}
+
+	return c.keycloakClient.DeleteRole(accessToken, realmName, roleID)
```
fperot74, line 2171: Can't we avoid calling GetRole and use DeleteRole return code to know if we might return a 404 or 200 status?

fperot74 (approved):

pkg/management/component.go
basbeu: The call to GetRole is there to ensure that we are deleting a client role and not a "realm" role. After discussing with @lagess, it is the best solution that we found.

harture (approved):

PR #404, opened 2022-06-17
[CLOUTRUST-4107] Use emailToValidate and phoneNumberToValidate

fperot74:


basbeu (approved):

pkg/management/component.go
```diff
@@ -585,17 +585,25 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
...
-	if user.Email.Defined && (user.Email.Value == nil || c.isUpdated(user.Email.Value, oldUserKc.Email)) {
-		var verified = false
-		user.EmailVerified = &verified
+	if user.Email.Defined {
+		if user.Email.Value == nil {
+			var verified = false
+			user.EmailVerified = &verified
+		} else if c.isUpdated(user.Email.Value, oldUserKc.Email) {
+			if err = c.checkEmailNotUsed(ctx, realmName, userID, *user.Email.Value); err != nil {
+				c.logger.Info(ctx, "msg", "Can't update email with email of an already existing account", "email", *user.Email.Value)
```
harture, line 594: To avoid that support leak the info avout email already in use, we could allow to set it but send an email which state the email is already in use. It would use the same logic in account and management.

pkg/management/component.go
```diff
@@ -585,17 +585,25 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
...
+		}
 	}
 
 	// when the phone number changes, set the PhoneNumberVerified to false
-	if user.PhoneNumber.Defined && (user.PhoneNumber.Value == nil || c.isUpdated(user.PhoneNumber.Value, oldUserKc.GetAttributeString(constants.AttrbPhoneNumber))) {
-		var verified = false
-		user.PhoneNumberVerified = &verified
+	if user.PhoneNumber.Defined {
 		if user.PhoneNumber.Value == nil {
 			removeAttributes = append(removeAttributes, constants.AttrbPhoneNumber, constants.AttrbPhoneNumberVerified)
```
harture, line 604: Should we also remove phoneNumberToValidate ?

pkg/management/component.go
```diff
@@ -585,17 +585,25 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 	}
 
 	// when the email changes, set the EmailVerified to false
-	if user.Email.Defined && (user.Email.Value == nil || c.isUpdated(user.Email.Value, oldUserKc.Email)) {
-		var verified = false
-		user.EmailVerified = &verified
+	if user.Email.Defined {
+		if user.Email.Value == nil {
+			var verified = false
```
harture, line 590: Do we still need to remove the flag ?

pkg/management/component.go
```diff
@@ -624,6 +632,10 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 	)
 
 	userRep = api.ConvertUpdatableToKCUser(user)
+	if user.Email.Value != nil {
+		// Keep previous email value. Update is managed with emailToValidate
+		userRep.Email = oldUserKc.Email
+	}
 
 	// Merge the attributes coming from the old user representation and the updated user representation in order not to lose anything
 	var mergedAttributes = make(kc.Attributes)
```
harture, line 641: How we ensure the phoneNumber/email is not directly updated ?

harture (changes requested):

api/account/api.go
```diff
@@ -127,33 +129,27 @@ func ConvertToAPIAccount(ctx context.Context, userKc kc.UserRepresentation, logg
...
+	userRep.EmailToValidate = userKc.GetAttributeString(constants.AttrbEmailToValidate)
 	userRep.FirstName = userKc.FirstName
 	userRep.LastName = userKc.LastName
+	userRep.PhoneNumber = userKc.GetAttributeString(constants.AttrbPhoneNumber)
+	userRep.PhoneNumberToValidate = userKc.GetAttributeString(constants.AttrbPhoneNumberToValidate)
+	userRep.Gender = userKc.GetAttributeString(constants.AttrbGender)
+	userRep.Locale = userKc.GetAttributeString(constants.AttrbLocale)
+	userRep.BusinessID = userKc.GetAttributeString(constants.AttrbBusinessID)
 
-	if value := userKc.GetAttributeString(constants.AttrbPhoneNumber); value != nil {
```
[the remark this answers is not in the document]
fperot74: Because we don't need to check if the attribute is there or not... everything is done in GetAttributeString (already returns nil if attribute is missing) and userRep has just been created so the default value is already nil

api/account/swagger-api_account.yaml
```diff
@@ -214,6 +214,9 @@ components:
         emailVerified:
           type: boolean
           description: only returned by /account
+        emailToValidate:
+          type: string
+          description: Read only. Email requested to be the email of the user but not yet validated
```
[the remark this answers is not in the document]
fperot74: The value is not taken into account if you provide it to update an account. This field only make sense in responses, not in requests

api/account/swagger-api_account.yaml
```diff
@@ -226,6 +229,9 @@ components:
         phoneNumberVerified:
           type: boolean
           description: only returned by /account
+        phoneNumberToValidate:
+          type: string
+          description: Read only. Phone number requested to be the phone number of the user but not yet validated
```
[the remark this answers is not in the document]
fperot74: same

pkg/account/component.go
```diff
@@ -230,39 +230,34 @@ func (c *component) UpdateAccount(ctx context.Context, user api.UpdatableAccount
...
 	if isUpdated(user.Email, oldUserKc.Email) {
-		var verified = false
-		emailVerified = &verified
-		if oldUserKc.Email != nil && len(*oldUserKc.Email) > 0 {
-			prevEmail = oldUserKc.Email
+		if err = c.checkEmailNotUsed(ctx, realm, userID, *user.Email); err != nil {
+			c.logger.Info(ctx, "msg", "Can't update email with email of an already existing account", "email", *user.Email)
+			return err
 		}
+		prevEmail = oldUserKc.Email
```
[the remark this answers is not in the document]
fperot74: Why?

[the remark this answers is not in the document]
harture: I cannot find the line where we are using it

pkg/account/component.go
```diff
@@ -319,6 +314,20 @@ func (c *component) UpdateAccount(ctx context.Context, user api.UpdatableAccount
 	return err
 }
 
+func (c *component) checkEmailNotUsed(ctx context.Context, realm string, userID string, email string) error {
```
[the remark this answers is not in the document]
fperot74: Repushed...
keycloak-rest-api-extensions will intercept ct-verify-email requests and, I email is already in use, will send a mail to tell that the change can't be validated (also have changes in cloudtrust-base-theme and custom-flows)

pkg/management/component.go
fperot74: In this case, we are explicitly asking to remove the email address... There is nothing to validate we can directly take it into account.
Alternative question would be: do we want to remove the email address? (I guess so... If we coded this, may be some customer asked it us)

pkg/management/component.go
fperot74: Good point... do we want to remove validated phone number or just the not-yet-validated one?

pkg/account/component.go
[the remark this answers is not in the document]
harture: Je crois qu'il faut que je m'achète des lunettes :)

pkg/management/component.go
fperot74: Repushed as discussed together

pkg/management/component.go
fperot74: Scenario seen together

harture (approved):

PR #411, opened 2022-07-15
[CLOUDTRUST-4422] Add back url parameters

basbeu:


fperot74 (approved):

harture (commented):
If I remeber correctly, we should have somewhere a default value. When we create a realm and we never persisted a RealmCustomConfiguration which value will be returned ? An empty string ? nil ?
Also what will happen for our existing client after the merge of this new feature ?

basbeu:
> If I remeber correctly, we should have somewhere a default value. When we create a realm and we never persisted a RealmCustomConfiguration which value will be returned ? An empty string ? nil ? Also what will happen for our existing client after the merge of this new feature ?

For now I did not put a default value, we have indeed a mechanism to do that. I am not sure about which default value to put.
In the current implmentation: for the existing clients, the field will be nil in the go struct.  Then  it will be omitted in the structure. 

I thought this way :
If we do not have any backURL defined, it is better to not have any.
For now, client did not use that so it should not break anything. 
For people that need it, we can just set this param and it will be present. 
But I do not know if this strategy is not good for the front end.

harture:
> > If I remeber correctly, we should have somewhere a default value. When we create a realm and we never persisted a RealmCustomConfiguration which value will be returned ? An empty string ? nil ? Also what will happen for our existing client after the merge of this new feature ?
> 
> For now I did not put a default value, we have indeed a mechanism to do that. I am not sure about which default value to put. In the current implmentation: for the existing clients, the field will be nil in the go struct. Then it will be omitted in the structure.
> 
> I thought this way : If we do not have any backURL defined, it is better to not have any. For now, client did not use that so it should not break anything. For people that need it, we can just set this param and it will be present. But I do not know if this strategy is not good for the front end.

That's fine for me, we just have to ensure MOB will manage correctly nil value in its system on front side

harture (approved):

api/management/api.go
```diff
@@ -207,6 +207,7 @@ type RealmCustomConfiguration struct {
 	OnboardingClientID                  *string   `json:"onboarding_client_id"`
 	SelfRegisterGroupNames              *[]string `json:"self_register_group_names"`
 	BarcodeType                         *string   `json:"barcode_type"`
+	BackURL                             *string   `json:"back_url,omitempty"`
```
harture, line 210: Might be better to name it allowed_back_url to make it more comprehensive of the exact role of it.

harture (approved):

basbeu:
> > > If I remeber correctly, we should have somewhere a default value. When we create a realm and we never persisted a RealmCustomConfiguration which value will be returned ? An empty string ? nil ? Also what will happen for our existing client after the merge of this new feature ?
> > 
> > 
> > For now I did not put a default value, we have indeed a mechanism to do that. I am not sure about which default value to put. In the current implmentation: for the existing clients, the field will be nil in the go struct. Then it will be omitted in the structure.
> > I thought this way : If we do not have any backURL defined, it is better to not have any. For now, client did not use that so it should not break anything. For people that need it, we can just set this param and it will be present. But I do not know if this strategy is not good for the front end.
> 
> That's fine for me, we just have to ensure MOB will manage correctly nil value in its system on front side

I checked with the front-end team and it is ok for them to not have the field when the configuration is not present

api/management/api.go
basbeu: Done

PR #412, opened 2022-07-15
[CLOUDTRUST-4417] Adapt GLN check

basbeu:


internal/business/gln.go
```diff
@@ -101,6 +101,7 @@ func (v *glnVerifier) ValidateGLN(firstName, lastName, gln string) error {
 }
 
 func (v *glnVerifier) compare(glnPerson GlnPerson, firstName, lastName, gln string) bool {
-	return glnPerson.Number != nil && glnPerson.FirstName != nil && glnPerson.LastName != nil &&
-		gln == *glnPerson.Number && firstName == *glnPerson.FirstName && lastName == *glnPerson.LastName
+	return glnPerson.Number != nil && gln == *glnPerson.Number &&
+		((glnPerson.FirstName != nil && firstName == *glnPerson.FirstName) || (glnPerson.FirstName == nil && firstName == "")) &&
```
fperot74, line 105: According to JIRA ticket, only name2 (lastName) can be nil/empty

internal/business/gln_test.go
```diff
@@ -81,6 +95,16 @@ func TestGln(t *testing.T) {
 		assert.Nil(t, err)
 	})
 
+	t.Run("Matching GLN, firstname nil", func(t *testing.T) {
```
fperot74, line 98: FirstName (name1) should not be empty

internal/business/gln.go
harture: We do not have a clear specification regarding constraint on Name1 and Name2. 
I don't see any possible issue to consider that Name1 might also be empty, even if it seems very unlikely. Do you see one ?

harture (approved):

internal/business/gln.go
fperot74: Just the one in the ticket :

> Juridical persons have Name1 and Name2 instead of Firstname and lastname. Name2 is optional and might be empty which is not supported by our current implementation. We want to adapt it to also support such scenario.

PR #419, opened 2022-08-23
[CLOUDTRUST-3326] Accreditation delivery & removal

basbeu:


internal/keycloakb/accountingclient.go
```diff
@@ -27,6 +27,7 @@ type accountingClient struct {
 // HTTPClient interface
 type HTTPClient interface {
 	Get(data interface{}, plugins ...plugin.Plugin) error
+	Post(data interface{}, plugins ...plugin.Plugin) (string, error)
```
fperot74, line 30: May be I added this but did not use... Could you try to remove it?

internal/keycloakb/accreditationsmodule.go
```diff
@@ -38,7 +33,7 @@ type AccredsKeycloakClient interface {
 
 // AdminConfigurationDBModule interface
 type AdminConfigurationDBModule interface {
-	GetAdminConfiguration(context.Context, string) (configuration.RealmAdminConfiguration, error)
+	//GetAdminConfiguration(context.Context, string) (configuration.RealmAdminConfiguration, error)
```
fperot74, line 36: Does it mean that this interface is not used anymore? (or else, it is an empty interface)

fperot74 (approved):
@harture , as we both worked on this PR with @basbeu, your review is mandatory

internal/keycloakb/accountingclient.go
basbeu: Good point ! Removed

internal/keycloakb/accreditationsmodule.go
basbeu: Good catch. I missed it during code cleaning

pkg/management/component.go
```diff
@@ -673,27 +689,27 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 	}
 
 	if userInfosUpdated {
-		if keycloakb.IsUpdated(user.BirthLocation, oldDbUser.BirthLocation) {
+		if fieldsComparator.IsAnyFieldUpdated(fields.BirthLocation) {
```
harture, line 692: Do we have a reason to not use CompareValues() which would perform the update directly ?

pkg/management/component.go
```diff
@@ -552,21 +566,14 @@ func (c *component) GetUser(ctx context.Context, realmName, userID string) (api.
...
 }
 
-func (c *component) isUpdated(newValue, oldValue *string) bool {
-	if oldValue == nil {
-		return newValue != nil
-	}
-	return newValue != nil && *oldValue != *newValue
-}
-
 func (c *component) UpdateUser(ctx context.Context, realmName, userID string, user api.UpdatableUserRepresentation) error {
```
harture, line 577: I have the feeling this method could now be simplified, to discuss

pkg/account/component.go
```diff
@@ -223,32 +236,43 @@ func (c *component) UpdateAccount(ctx context.Context, user api.UpdatableAccount
...
 		actions = append(actions, ActionVerifyEmail)
+		profileChangesCount-- //
 	}
 
 	// manage phone number change
-	if isUpdated(user.PhoneNumber, oldUserKc.GetAttributeString(constants.AttrbPhoneNumber)) {
+	if fieldsComparator.IsAnyFieldUpdated(fields.PhoneNumber) {
 		oldUserKc.SetAttributeString(constants.AttrbPhoneNumberToValidate, *user.PhoneNumber)
 		actions = append(actions, ActionVerifyPhoneNumber)
 	}
```
harture, line 278: Could we use the new FieldsComparator to perform the update of the fields ?

pkg/account/component.go
```diff
@@ -223,32 +236,43 @@ func (c *component) UpdateAccount(ctx context.Context, user api.UpdatableAccount
...
 		oldUserKc.SetAttributeString(constants.AttrbEmailToValidate, *user.Email)
 		actions = append(actions, ActionVerifyEmail)
+		profileChangesCount-- //
 	}
 
 	// manage phone number change
-	if isUpdated(user.PhoneNumber, oldUserKc.GetAttributeString(constants.AttrbPhoneNumber)) {
+	if fieldsComparator.IsAnyFieldUpdated(fields.PhoneNumber) {
 		oldUserKc.SetAttributeString(constants.AttrbPhoneNumberToValidate, *user.PhoneNumber)
 		actions = append(actions, ActionVerifyPhoneNumber)
```
harture, line 277: profileChangesCount-- missing ?

harture (changes requested):

pkg/validation/component.go
```diff
@@ -197,7 +231,7 @@ func (c *component) updateUserDatabase(ctx context.Context, realmName, userID st
 	}
 
 	if existingUser, err := c.usersDBModule.GetUserDetails(ctx, realmName, userID); err == nil {
-		shouldRevokeAccreditations = user.HasUpdateOfAccreditationDependantInformationDB(existingUser)
+		_ = user.HasDBChanges(fc, existingUser)
```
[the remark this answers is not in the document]
basbeu: Fixed

pkg/validation/component.go
```diff
@@ -208,26 +242,36 @@ func (c *component) updateUserDatabase(ctx context.Context, realmName, userID st
...
+func (c *component) updateUserKeycloak(validationCtx *validationContext, user api.UserRepresentation, fc fields.FieldsComparator) error {
 	var kcUser, err = c.loadKeycloakUserCtx(validationCtx)
 	if err != nil {
 		return err
 	}
 	keycloakb.ConvertLegacyAttribute(kcUser)
-	if !shouldRevokeAccreditations || user.HasUpdateOfAccreditationDependantInformationKC(kcUser) {
-		shouldRevokeAccreditations = true
-	}
+	_ = user.HasKCChanges(fc, kcUser)
```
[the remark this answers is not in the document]
basbeu: Fixed

internal/keycloakb/accreditationsmodule.go
```diff
@@ -55,162 +39,174 @@ type AccreditationRepresentation struct {
...
+}
+
+func NewAccreditationsProcessor(accreditations []string) (AccreditationsProcessor, error) {
+	var res = accredsProcessor{accreditations: make(map[string][]AccreditationRepresentation)}
+	var bestEffortError error
+
+	for _, accred := range accreditations {
+		var newAccred AccreditationRepresentation
+		if err := json.Unmarshal([]byte(accred), &newAccred); err != nil {
+			// Will return an error but try to go on processing. User will decide if he wants to process accreditations anyway or not
```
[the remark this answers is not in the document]
basbeu: The user is a poor wording. It is meant to be the caller. It is up to the caller of the method to treat the error.

pkg/management/component.go
basbeu: CompareValues is meant to build add a values to compare in the comparator. I think this is the right method to call. For more clarity, I will call IsFieldUpdated

pkg/account/component.go
basbeu: No, this variable is only used to check if we have to send an email for a change of the profile. But we do not want to take into account the email. But in case of a phone number change, we want to send the profile email change

pkg/account/component.go
basbeu: The FieldsComparator does not have this feature. We could imagine a method to do that, but I am not sure that is good idea

pkg/management/component.go
basbeu: Hum not trivial to do, we can maybe split it in small submethods

internal/keycloakb/accreditationsmodule.go
```diff
@@ -55,162 +39,174 @@ type AccreditationRepresentation struct {
...
+	return &accreditationsEvaluator{
+		accredsSvcClient: accredsSvcClient,
+		logger:           logger,
 	}
-	return newAccreds, nil
 }
 
-func (am *accredsModule) convertDurationToDate(ctx context.Context, validity string) (*string, error) {
-	var expiryDate, err = validation.AddLargeDurationE(time.Now(), validity)
+func (ae *accreditationsEvaluator) EvaluateAccreditations(ctx context.Context, realmName string, userID string, fieldsComparator fields.FieldsComparator, currentAccreds []string) ([]string, error) {
```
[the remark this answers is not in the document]
basbeu: I think this not a role for the accreditation client !
So either we put it in a module with the advantages (as I did first) : 
- ease the testing as we can mock it in the component and test it in module
- code put in a smaller file

Or we have to put this code at least twice in account and management package. We can modularize in both places in a submethod but the method is still in a +500 lines file.

I propose a version where the code is directly in methods

pkg/account/component.go
harture: Sorry, I do not understand it ... and I have the feeling that it is becoming too much complicated to maintain it without risk of introduction of regressions.

harture (approved):

pkg/account/component.go
fperot74: There are specific email sent for email changes... So for the "profile changes email", we dont take email changes into account ; that's why we decrement profileChangesCount.
For phone number, there is no specific email sent... so phone number do need to be taken into account in profileChangesCount

PR #428, opened 2022-09-08
Fix context-key loading

fperot74:


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -968,7 +968,9 @@ func main() {
 		var onboardingModule = keycloakb.NewOnboardingModule(keycloakClient, keycloakConfig.URIProvider, usersDBModule, registerInactiveLockDuration, onboardingRealmOverrides, registerLogger)
 
 		// context keys
-		contextKeyManager, err := keycloakb.MakeContextKeyManager(c.Get(cfgContextKeys))
+		contextKeyManager, err := keycloakb.MakeContextKeyManager(func(config interface{}) error {
```
harture, line 971: could we directly provide the result of this func instead of doing it in a functional style ?

harture (approved):

PR #434, opened 2022-10-11
[CLOUDTRUST-4566] Send Verify phone number email only if phone number is already set

fperot74:


pkg/management/component.go
```diff
@@ -624,7 +624,9 @@ func (c *component) UpdateUser(ctx context.Context, realmName, userID string, us
 			removeAttributes = append(removeAttributes, constants.AttrbEmailToValidate)
 		} else if fieldsComparator.IsAnyFieldUpdated(fields.Email) {
 			oldUserKc.SetAttributeString(constants.AttrbEmailToValidate, *user.Email.Value)
-			actions = append(actions, actionVerifyEmail)
+			if len(oldUserKc.GetFieldValues(fields.PhoneNumber)) > 0 {
```
basbeu, line 627: Does it really solve the problem ? We still send an email at line 637 ?

And I do not really see the point to change the behavior of the email changed for the specific problems we want to solve

fperot74: Fixed

harture (approved):

PR #437, opened 2022-10-18
[CLOUDTRUST-4511] Send multiple proofs for physical identification

basbeu:


pkg/kyc/component.go
```diff
@@ -663,3 +682,27 @@ func (c *component) sendSmsCodeGeneric(ctx context.Context, accessToken string,
...
 	return *smsCodeKc.Code, err
 }
+
+func (c *component) zipAttachments(attachments []apikyc.AttachmentRepresentation) ([]byte, error) {
+	buf := new(bytes.Buffer)
+
+	w := zip.NewWriter(buf)
+
+	for _, attachment := range attachments {
+		f, err := w.Create(*attachment.Filename)
```
fperot74, line 692: I suggest to set a creation date when adding files to avoid having an archive with files from 1970.

pkg/kyc/component.go
```diff
@@ -663,3 +682,27 @@ func (c *component) sendSmsCodeGeneric(ctx context.Context, accessToken string,
 
 	return *smsCodeKc.Code, err
 }
+
+func (c *component) zipAttachments(attachments []apikyc.AttachmentRepresentation) ([]byte, error) {
```
fperot74, line 686: I would create an external tool in /internal/xxx responsible of ZIP creation.
This way, you will be able to have unit tests dedicated to the zip creation... and avoid having to mock-up zip processing here.
(ex: call a NewZipWriter in NewComponent and override the value in unit tests)

fperot74 (approved):

pkg/kyc/component.go
basbeu: Why not, but we actually do not need to mock anything of this part to test it

pkg/kyc/component.go
fperot74: Ok... I did not see that you were not using the same library as we did in papercard...
https://bitbucket.svc.elca.ch/projects/CLOUDTRUST/repos/2019.oc.paper-card-service/browse/internal/papercard/zipcreator.go#19
So may be archive/zip does not have this date issue

pkg/kyc/component.go
fperot74: Of course we don't need to mock anything... there is no unit test covering this part.

basbeu: I did a dedicated test : TestZipAttachments

harture: Maybe Francis was thinking about a test of the public method with multiple attachments. For now, we have a test which covers the private method in charge of the ZIP but not a one which ensure a ZIP is created if you provide multiple attachments.

harture (approved):

PR #438, opened 2022-11-04
[CLOUDTRUST-4593] Migrate users details to keycloak attributes

basbeu:


api/kyc/api.go
```diff
@@ -164,6 +159,12 @@ func (u *UserRepresentation) ImportFromKeycloak(ctx context.Context, kcUser *kc.
 	var accreditations = u.Accreditations
 	var locale = u.Locale
 	var businessID = u.BusinessID
+	var birthLocation = u.BirthLocation
```
fperot74, line 162: Readability is more and more heavy... May be we could write something like:
var birthLocation = defaultIfNil(kcUser.GetAttributeString(constants.AttrbBirthLocation), u.BirthLocation)

It would divide number of lines of code per 4. (and reduce number of branches)

api/validation/api.go
```diff
@@ -146,6 +151,12 @@ func (u *UserRepresentation) ImportFromKeycloak(kcUser kc.UserRepresentation) {
 	var phoneNumberVerified = u.PhoneNumberVerified
 	var gender = u.Gender
 	var birthdate = u.BirthDate
+	var birthLocation = u.BirthLocation
```
fperot74, line 154: Same here (see comment  with defaultIfNil)

fperot74 (approved):

harture (approved):

api/kyc/api.go
basbeu: Good idea, this method is indeed too long

basbeu: done

PR #440, opened 2022-11-22
[CLOUDTRUST-4637] Get users profile

fperot74:
Requires :
* cloudtrust/common-service#113
* cloudtrust/keycloak-client#106


api/account/swagger-api_account.yaml
```diff
@@ -315,6 +327,65 @@ components:
...
+          $ref: '#/components/schemas/ProfileAttributeValidations'
+    ProfileAttributeValidations:
+      description: >
+        Here is a list of known validators :
+        email: no parameters
+        length: min/max (can be integer or string), trim-disabled (boolean as string)
+        integer: min/max (integer as string)
+        double: min/max (integer as string)
+        options: options (array of allowed values)
+        pattern: pattern (regex), error-message (string)
```
wormsparty, line 364: Ca sera quoi le format des Regexp? Je vois déjà venir que ça ne sera pas forcément compatible entre Go et Javascript...

fperot74: Exemple :
```
        "pattern": {
          "pattern": "^\\w+$",
          "error-message": "error.message.key"
        },
```

pkg/account/component.go
```diff
@@ -31,6 +32,8 @@ const (
 	emailSubjectUpdatedEmail     = "notifEmailChangeSubject"
 	emailTemplateUpdatedProfile  = "notif-profile-change.ftl"
 	emailSubjectUpdatedProfile   = "notifProfileChangeSubject"
+
+	frontendType = "sse"
```
basbeu, line 36: Should this constant not be defined at the endpoint level ? 
Minor thing : I also challenge the name, sse refers to self-service that is a webapp. This api can be used by other frontend. Is it really a type ?

pkg/management/component.go
```diff
@@ -33,6 +34,8 @@ const (
 
 	actionVerifyEmail       = "ct-verify-email"
 	actionVerifyPhoneNumber = "mobilephone-validation"
+
+	frontendType = "bo"
```
basbeu, line 38: Should not be defined at endpoint level ?

pkg/register/component.go
```diff
@@ -5,16 +5,20 @@ import (
...
+	apicommon "github.com/cloudtrust/keycloak-bridge/api/common"
 	apiregister "github.com/cloudtrust/keycloak-bridge/api/register"
 	"github.com/cloudtrust/keycloak-bridge/internal/constants"
 	"github.com/cloudtrust/keycloak-bridge/internal/keycloakb"
 	kc "github.com/cloudtrust/keycloak-client/v2"
+	"github.com/cloudtrust/keycloak-client/v2/toolbox"
+)
+
+const (
+	frontendType = "register"
```
basbeu, line 21: Here again, I think this should be put in the endpoint

pkg/register/validation.go
```diff
@@ -0,0 +1,68 @@
+package register
```
basbeu, line 1: Why here and not in the endpoint as it has been done for the others api ?

internal/keycloakb/fields_validation.go
```diff
@@ -0,0 +1,44 @@
+package keycloakb
```
basbeu, line 1: I do not find any reference of this tool in the code. It is maybe a first version of the profile cache that is now in the keycloak-client

basbeu (changes requested):

api/account/api.go
```diff
@@ -209,48 +211,94 @@ func ConvertToKCUser(user UpdatableAccountRepresentation) kc.UserRepresentation
...
-		ValidateParameterIn(constants.IDDocumentType, user.IDDocumentType, constants.AllowedDocumentTypes, false).
-		ValidateParameterRegExp(constants.IDDocumentNumber, user.IDDocumentNumber, constants.RegExpIDDocumentNumber, false).
-		ValidateParameterLength(constants.IDDocumentNumber, user.IDDocumentNumber, 1, 50, false).
-		ValidateParameterDateMultipleLayout(constants.IDDocumentExpiration, user.IDDocumentExpiration, constants.SupportedDateLayouts, false).
-		ValidateParameterRegExp(constants.IDDocumentCountry, user.IDDocumentCountry, constants.RegExpCountryCode, false).
-		Status()
+// GetField is used to validate a user against a UserProfile
+func (user *UpdatableAccountRepresentation) GetField(field string) interface{} {
+	switch field {
+	//case fields.Username.Key():
```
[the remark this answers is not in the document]
fperot74: It was waiting for the Username field to be created... it is now fixed

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+
+// AttributesToApi converts a KC profile attribute to its API version
+func AttributesToApi(attrbs []kc.ProfileAttrbRepresentation, frontend string, requesterType string) []ProfileAttributeRepresentation {
+	if attrbs == nil {
+		return nil
+	}
+	if len(attrbs) == 0 {
+		return make([]ProfileAttributeRepresentation, 0)
+	}
+	var res []ProfileAttributeRepresentation
```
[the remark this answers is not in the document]
fperot74: It would be a long story to explain this...
Anyway, I don't have time to investigate, I'm following your suggestion.

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+			res = append(res, *newValue)
+		}
+	}
+	return res
+}
+
+func AttributeToApi(attrb kc.ProfileAttrbRepresentation, frontend string, requesterType string) *ProfileAttributeRepresentation {
+	if attrb.Permissions != nil && !validation.IsStringInSlice(attrb.Permissions.Edit, requesterType) {
+		// User has no permission to edit this field
+		// Register should not be aware of its existence
```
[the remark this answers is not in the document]
fperot74: I first wrote this for the register component the decided to make it a common tool.
Fixed

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+	GetField(name string) interface{}
+	SetField(name string, value interface{})
+}
+
+// ValidateUser against a given profile
+func ValidateUser(profile kc.UserProfileRepresentation, input ContainsFields, frontendType string, requesterType string, checkMandatory bool) error {
+	for _, attrb := range profile.Attributes {
+		if !attrb.IsAnnotationTrue(frontendType) {
+			// Attribute is not supposed to be provided for this frontend type
+			input.SetField(*attrb.Name, nil)
```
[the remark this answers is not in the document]
fperot74: What Keycloak would do with an unnamed attribute? It would not be usable.

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+	return nil
+}
+
+func validateAttributeEmail(attrb kc.ProfileAttrbRepresentation, validator kc.ProfileAttrValidatorRepresentation, value interface{}) error {
+	switch v := value.(type) {
+	case string:
+		return validateRegexWithString(attrb, constants.RegExpEmail, v)
+	case *string:
+		return validateRegexWithString(attrb, constants.RegExpEmail, *v)
```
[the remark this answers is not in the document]
fperot74: No... Function is called in ValidateUser(...) only if v is not nil

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+			return cerrors.CreateInternalServerError("unknownValidator." + key)
+		}
+	}
+	return nil
+}
+
+func validateAttributeEmail(attrb kc.ProfileAttrbRepresentation, validator kc.ProfileAttrValidatorRepresentation, value interface{}) error {
+	switch v := value.(type) {
+	case string:
+		return validateRegexWithString(attrb, constants.RegExpEmail, v)
```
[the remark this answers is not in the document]
fperot74: Their regexp is awful... but anyway, it will be the final regex used by KC... fixed!

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+	switch v := value.(type) {
+	case *string:
+		return v
+	}
+	var res = fmt.Sprintf("%v", value)
+	return &res
+}
+
+func toInt(value interface{}, defaultValue int) int {
```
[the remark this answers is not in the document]
fperot74: Attribute validation functions starting with a lowercase character are called only if value is not nil

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+	case string:
+		return atoi(v, defaultValue)
+	case *string:
+		return atoi(*v, defaultValue)
+	default:
+		return defaultValue
+	}
+}
+
+func toFloat(value interface{}, defaultValue float64) float64 {
```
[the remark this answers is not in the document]
fperot74: Not called with nil value

api/kyc/api.go
```diff
@@ -252,6 +242,101 @@ func (u *UserRepresentation) Validate(everythingOptional bool) error {
...
+	return value
+}
+
+// GetField is used to validate a user against a UserProfile
+func (user *UserRepresentation) GetField(field string) interface{} {
+	switch field {
+	//case fields.Username.Key():
+	//	return user.Username
+	case fields.Email.Key():
+		return ifNotNil(user.Email)
```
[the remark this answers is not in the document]
fperot74: That's a mess... You sometimes have nil values which are not recognized as nil values...
I saw this during writing of KYC tests... may be it is missing in account

pkg/account/component.go
fperot74: If one day, a client wants to use this method to implement his own interface, I might not need our self-service... So we will be able to define his configuration using the "sse" value.
If the client wants to have our self-service AND its own self-service with different attributes, then we will have to find a solution... but I don't think this will be done in the future. (and if he wants to implements its own self-service, I'm not sure he will have to have a dynamic configuration)

pkg/management/component.go
fperot74: Already answered for account/component.go

pkg/register/component.go
fperot74: Already answered for account/component.go

pkg/register/validation.go
fperot74: A test we decided to do with @harture. If this new middleware is validated, it will be a very long job to move all necessary code for management interface. That's why we decided to do this only in register.

basbeu: Okay got it. My opinion :  I do not think we need to do a middleware for validation and think we can remove it and make register the same as the others.

pkg/account/component.go
basbeu: I think you did not really answer my first point : I do not really see the point of defining this constant in component.go and endpoint.go. I think we will always define this constant in endpoint (example KYC). So I think we can only define it in endpoint.go.


Maybe I should rephrase the second point : I challenged the name, because we are in a api called account and now we are binding it with the frontendType. In my point of view, the bridge is not aware of the front-end and is do not need to be aware of  which frontend is calling him. So why putting a parameter called sse ? I think this is misleading and the parameter should have the name of the api.

api/kyc/api.go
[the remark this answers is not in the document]
basbeu: Your explanation sounds really strange to me. Can you provide an example ? or more context ? 
If it is really the case, it is something that can be dangerous and that we need to be aware of.

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+		// Should not happen if correctly implemented
+		return cerrors.CreateInternalServerError("unknownInputType")
+	}
+}
+
+func validateRegexWithString(attrb kc.ProfileAttrbRepresentation, regex string, value string) error {
+	// Java: \p{L} Golang: \p{Lu}\p{Ll}
+	regex = strings.ReplaceAll(regex, `\h`, `\s`)
+	regex = strings.ReplaceAll(regex, `\v`, `\s`)
+	regex = strings.ReplaceAll(regex, `\p{Cntrl}`, `\x00-\x1F\x7F`)
```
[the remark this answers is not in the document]
basbeu: So may I ask what is the purpose of this case ? and what is the use case ? How did you end up with \x00-\x1F\x7F ?

api/common/api-profile.go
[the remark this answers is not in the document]
basbeu: Yes, I think this case will not happen but we know that things that should never happen often happen with a scenario we did not think about.

api/common/api-profile.go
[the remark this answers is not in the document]
basbeu: okay, they are private so it is acceptable. Should we add a comment to the method to precise the assumption of not-nil interface for the value param ?

basbeu (changes requested):
There is a failed unit test {"Time":"2022-12-22T16:07:14.896950846+01:00","Action":"output","Package":"github.com/cloudtrust/keycloak-bridge/api/register","Test":"TestGetField","Output":"--- FAIL: TestGetField (0.00s)\n"}

pkg/account/component.go
fperot74: I can move declaration in endpoint... it was created where it was used, in component (it was first used only there). Now endpoint also use it.

I don't really care about the name (you can make suggestions... frontendName?)... I think the most important is the value...
The attribute management is clearly designed to be adapted for a specific context. I just wanted to help integrators to be able to configure realms the most simply as possible: management is too long and you can easily swap to letters. How should we write: self service, self-service, self_service ? To avoid doubt, I used sse.

api/kyc/api.go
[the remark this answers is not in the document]
fperot74: When you start playing with interface{} and reflection, you sometimes have a nil value and when you check value==nil, result is false. We already had such an issue with validation: https://github.com/cloudtrust/common-service/blob/master/validation/validator.go#L84

api/common/api-profile.go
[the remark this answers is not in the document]
fperot74: Added comments

api/kyc/api.go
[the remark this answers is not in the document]
basbeu: Okay interesting ! So this case is covered by your unit tests ? I think it is good to have it to avoid regression in future changes.

pkg/account/component.go
basbeu: if we want to put : sse, bo ... as values frontendType is ok for me.

I would suggest to have apiName with the name of the api. I think it is not good to over simplify our system for integrator. It is only my personal point of view. If you do not agree and @harture is fine with it, it is okay to keep as it is now.

basbeu (approved):

pkg/kyc/component.go
```diff
@@ -63,6 +63,11 @@ type AccreditationsServiceClient interface {
 	NotifyCheck(ctx context.Context, check accreditationsclient.CheckRepresentation) error
 }
 
+// UserProfileCache interface
+type UserProfileCache interface {
+	GetRealmUserProfile(ctx context.Context, realmName string) (kc.UserProfileRepresentation, error)
+}
+
 // Component is the register component interface.
 type Component interface {
```
harture, line 72: If we go with the idea of having apiNames instead of frontendTypes, it woudl make sense to have a user profile endpoint also for kyc API to be coherent and having full flexibility.

api/kyc/api.go
```diff
@@ -217,20 +219,8 @@ func defaultIfNil(value *string, defaultValue *string) *string {
 }
 
 // Validate checks the validity of the given User
-func (u *UserRepresentation) Validate(everythingOptional bool) error {
+func (u *UserRepresentation) Validate() error {
```
harture, line 222: I think we should drop this method as it should not be used anymore no ?

api/common/api-profile.go
```diff
@@ -0,0 +1,457 @@
...
+		} else if k == "max" {
+			var attributeValue = toInt(v, referenceValue-1) // Use a too low value by default to make the test fails
+			if referenceValue > attributeValue {
+				return cerrors.CreateBadRequestError(cerrors.MsgErrInvalidParam + "." + *attrb.Name)
+			}
+		}
+	}
+	return nil
+}
+
```
harture, line 361: Maybe all this generic conversion methods bellow could be move to commons ?
It would allow us to reuse it and reduce the size of this big api-profile, what do you think ?

api/common/api-profile.go
```diff
@@ -0,0 +1,457 @@
+package apicommon
```
harture, line 1: To discuss, I have a bit the feelign this file could be splitted into 2 pieces.
There is all the definitions of the Profile strcuture and conversions methods which are API related.
There is also methods and logic for the internal validation. We could move all of this into a module/tool in /internal. It would avoid to have for exemple the default-profile into /api whereas it is something which is used very deep by UserProfileCache system.

api/common/api-profile.go
```diff
@@ -0,0 +1,457 @@
...
+// ValidateUser against a given profile
+func ValidateUser(profile kc.UserProfileRepresentation, input ContainsFields, frontendType string, requesterType string, checkMandatory bool) error {
+	for _, attrb := range profile.Attributes {
+		if !attrb.IsAnnotationTrue(frontendType) {
+			// Attribute is not supposed to be provided for this frontend type
+			input.SetField(*attrb.Name, nil)
+		} else {
+			var value = input.GetField(*attrb.Name)
+			if value == nil {
+				if checkMandatory && isAttributeRequired(attrb, requesterType) {
```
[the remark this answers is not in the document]
fperot74: In case you are in an update context, you may want to disable the "mandatory" checks as you know that provided nil values will keep the previous value (which has already been validated)

api/kyc/api.go
fperot74: We still need to check some values which are not going to be stored in KC (here, attachments)

api/common/api-profile.go
```diff
@@ -0,0 +1,441 @@
...
+			DisplayHeader:      group.DisplayHeader,
+			DisplayDescription: group.DisplayDescription,
+			Annotations:        group.Annotations,
+		})
+	}
+	return res
+}
+
+// DefaultProfile returns a default profile
+func DefaultProfile(realmName string) (kc.UserProfileRepresentation, error) {
```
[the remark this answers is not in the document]
fperot74: I would I preffered to have this in keycloak-client... but I consider keycloak-client as a golang client for Keycloak. If I move the default profile in keycloak-client, it would be a cloudtrust-specific default configuration, that's why I kept it here

harture (approved):

basbeu (approved):

api/common/api-profile.go
```diff
@@ -72,14 +83,20 @@ func AttributeToApi(attrb kc.ProfileAttrbRepresentation, frontend string) *Profi
...
 		DisplayName: attrb.DisplayName,
 		Group:       attrb.Group,
 		Required:    &required,
 		Validations: ValidationsToApi(attrb.Validations),
+		Annotations: AttributeAnnotationsToApi(attrb.Annotations),
 	}
 }
 
+func cleanUpName(name *string) *string {
+	var res = strings.ReplaceAll(*name, "ENC_", "")
```
basbeu, line 96: Test  TestAttributesToApi/Two_elements failed with 
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
```

basbeu (changes requested):

basbeu (approved):

PR #447, opened 2023-02-28
[CLOUDTRUST-4637] Move expected length from pattern to length validator

fperot74:


internal/profile/default-profile.json
```diff
@@ -95,9 +95,14 @@
 		  ]
 		},
 		"validations": {
-		  "pattern": {
-			"pattern": "^([\\p{Lu}\\p{Ll}][\\p{Lu}\\p{Ll} /\\\\\\.',-]{0,127})$"
-		  }
+			"pattern": {
```
basbeu, line 98: why do we change the regexp ? I could imagine that we let the original regexp and we  just add the length fields

basbeu (approved):

internal/profile/default-profile.json
fperot74: We don't want a message telling that the pattern is invalid and we can't know in which order patterns will be evaluated

PR #457, opened 2023-03-14
Update swagger for register interface

fperot74:


api/register/swagger-api_register.yaml
```diff
@@ -73,11 +92,11 @@ paths:
...
-  /register/config:
+  /register/realms/{realm}/user/profile:
     get:
       tags:
       - Configuration
-      summary: Gets configuration
+      summary: Gets the definition of user attributes from the realm specified in URL path
       parameters:
       - name: realm
         in: query
```
basbeu, line 102: path not query

basbeu (approved):

api/register/swagger-api_register.yaml
fperot74: > path not query

Copy/paste of errors present in the repo for 3 years...

PR #459, opened 2023-03-20
[CLOUDTRUST-4637] Allow validation to update some attributes in default user profile

fperot74:


basbeu:
I know we also send to the validation endpoint those fields : 
- email adress
- phonenumber
- local
To what I see, you did not change anything for those three fields. Are we sure it is correct ?

fperot74:
@basbeu, if you look at the code in *func computeUserUpdates()*, you'll see the fields which are taken from the IDnow payload

harture (approved):

basbeu (approved):

PR #465, opened 2023-04-12
[CLOUDTRUST-4637] Remove the required section of lastname/firstname

fperot74:


basbeu:
General question : here we remove the required section for lastname/firstname because of the management api. But it is maybe required for another api. How do we handle this scenario ?

harture (approved):

basbeu (approved):

fperot74:
> General question : here we remove the required section for lastname/firstname because of the management api. But it is maybe required for another api. How do we handle this scenario ?

The removed "required" section is a global requirement statement (Keycloak requirement). When we want to make an attribute required, we precise it in the configuration on a per interface basis using annotations (not used by keycloak, some kind of "user defined comment")

PR #469, opened 2023-06-09
[CLOUDTRUST-4985] Fix bug to avoid removing first/last names when not expected

fperot74:


api/account/api.go
```diff
@@ -181,23 +181,39 @@ func convertToAccreditations(ctx context.Context, values []string, logger keyclo
...
+	//userRep.Email = defaultStringPtr(user.Email, userRep.Email)
+	userRep.FirstName = defaultStringPtr(user.FirstName, userRep.FirstName)
+	userRep.LastName = defaultStringPtr(user.LastName, userRep.LastName)
 
 	var attributes = make(kc.Attributes)
-	attributes.SetStringWhenNotNil(constants.AttrbPhoneNumber, user.PhoneNumber)
+	if userRep.Attributes != nil {
+		attributes = *userRep.Attributes
+	}
+	// Don't set email as email and phoneNumber have specific validation rules
```
harture, line 203: Don't set phone number

api/account/api.go
```diff
@@ -181,23 +181,39 @@ func convertToAccreditations(ctx context.Context, values []string, logger keyclo
...
+}
 
-	userRep.Username = user.Username
-	userRep.Email = user.Email
-	userRep.FirstName = user.FirstName
-	userRep.LastName = user.LastName
+// MergeUser merge new values into existing KC user representation
+func MergeUser(userRep *kc.UserRepresentation, user UpdatableAccountRepresentation) {
+	userRep.Username = defaultStringPtr(user.Username, userRep.Username)
+	// Don't set email as email and phoneNumber have specific validation rules
```
harture, line 194: Ok, but where do you manage it now ?

api/account/api.go
harture: Where is it managed ?

api/account/api.go
```diff
@@ -181,23 +181,39 @@ func convertToAccreditations(ctx context.Context, values []string, logger keyclo
...
+	//userRep.Email = defaultStringPtr(user.Email, userRep.Email)
+	userRep.FirstName = defaultStringPtr(user.FirstName, userRep.FirstName)
+	userRep.LastName = defaultStringPtr(user.LastName, userRep.LastName)
 
 	var attributes = make(kc.Attributes)
```
harture, line 199: Why we don't use the duplicateAttribute method ?

api/management/api.go
```diff
@@ -298,15 +298,23 @@ type IdentityProviderRepresentation struct {
 type RequiredAction string
 
 func defaultString(actual *string, defaultValue string) *string {
```
harture, line 300: This method is a bit strange in its behavior. The output might be a reference to the initial string or not. I have a bad feeling about it.

harture (changes requested):

api/account/api.go
fperot74: > Don't set phone number

It is not set...

harture:
Yes. MY comment was to update your comment. It talks about email instead of phone number 😉

api/account/api.go
fperot74: Because it does not exist... and we are merging values, not duplicating them

api/management/api.go
fperot74: That's the same defaultString mechanism as the one in Apache common lang 3

api/account/api.go
fperot74: It is the responsibility of the caller (UpdateAccount in pkg/account)

basbeu: I do not understand the link between specific validation rules and that we do not merge them ! 
In my point of view a merge method should merge every fields.

basbeu: In my point of view it is really not a good idea !

basbeu (changes requested):

basbeu:
I saw that the PR is functional, but I think that the email and phone number should be merged !

api/account/api.go
fperot74: Repushed with renamed function

basbeu (approved):

harture (approved):

PR #472, opened 2023-07-07
[CLOUDTRUST-4941] Migration to Go 1.20

fperot74:


vloup (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cloudtrust/keycloak-bridge
 
-go 1.17
+go 1.20
```
basbeu, line 3: This is not useful : 
"The go directive sets the minimum version of Go required to use this module" , https://go.dev/ref/mod

Do we really want to update each component changing only this line ? I think if we do not need to do code updates we can only do the update in the CI side !

basbeu (changes requested):

PR #476, opened 2023-08-10
[CLOUDTRUST-5108] Compare trimmed data when checking GLN values

fperot74:


internal/business/gln_test.go
```diff
@@ -121,3 +121,21 @@ func TestGln(t *testing.T) {
 		time.Sleep(time.Second)
 	})
 }
+
+func TestCompare(t *testing.T) {
```
basbeu, line 125: test case with nil gln number is potentially missing : 
`GlnPerson{Number: nil, FirstName: ptr(" firstname "), LastName: ptr(" lastname ")}`

basbeu (approved):

internal/business/gln_test.go
fperot74: GLN is the identifier... you can't have identified someone if you don't have a GLN number.

PR #483, opened 2023-09-25
[CLOUDTRUST-5196] Add liveness endpoint binded to the AuditEventReporter

basbeu:


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -543,6 +544,10 @@ func main() {
 	healthChecker.AddDatabase("Archive RO", archiveRwDBConn, healthCheckCacheDuration)
 	healthChecker.AddHTTPEndpoints(c.GetStringMapString("healthcheck-endpoints"), httpTimeout, 200, healthCheckCacheDuration)
 
+	var livenessChecker = healthcheck.NewHealthChecker(keycloakb.ComponentName, logger)
```
fperot74, line 547: Couldn't we merge live endpoint with check endpoint ? Do we really need to have both at the same time ?

fperot74 (approved):

PR #486, opened 2023-10-20
[CLOUDTRUST-5220] Include details in execute actions event

basbeu:


harture (approved):

fperot74 (approved):

pkg/management/component.go
```diff
@@ -1172,7 +1172,7 @@ func (c *component) ExecuteActionsEmail(ctx context.Context, realmName string, u
 	values := append(paramKV, "required_actions", listActions)
 
 	details := map[string]string{}
-	for i := 0; i+1 < len(details); i += 2 {
+	for i := 0; i+1 < len(values); i += 2 {
```
harture, line 1175: Could we have a test that would have detected this mistake ?

basbeu: Yes, but not easily, we usually do not check the values of audit events

PR #487, opened 2023-10-20
[CLOUDTRUST-4933] Remove events, statistics and export package + remo…

basbeu:
…ve audit event mariadb


pkg/events/authorization.go
```diff
@@ -1,88 +0,0 @@
...
-		return &authorizationComponentMW{
-			authManager: authorizationManager,
-			logger:      logger,
-			next:        next,
-		}
-	}
-}
-
-func (c *authorizationComponentMW) GetActions(ctx context.Context) ([]api.ActionRepresentation, error) {
-	var action = security.EVGetActions.String()
```
fperot74, line 31: Actions might be removed from common-service? or this code was already moved in another service?

pkg/statistics/authorization.go
```diff
@@ -41,16 +41,6 @@ func (c *authorizationComponentMW) GetActions(ctx context.Context) ([]api.Action
 	return c.next.GetActions(ctx)
 }
 
-func (c *authorizationComponentMW) GetStatistics(ctx context.Context, realm string) (api.StatisticsRepresentation, error) {
-	var action = security.STGetStatistics.String()
```
fperot74, line 45: Same for these actions...

fperot74 (approved):

pkg/events/authorization.go
basbeu: Used in event service

pkg/statistics/authorization.go
basbeu: Same here

PR #497, opened 2024-04-04
[CLOUDTRUST-5142] Add events when accreditations are granted/revoked

fperot74:
Add events when an accreditation is granted/revoked
Rename credentialID detail to credential_id in events for consistency reason (credential_id is used in our Keycloak modules)


internal/keycloakb/accreditationsmodule.go
```diff
@@ -53,9 +63,9 @@ func RevokeAccreditations(kcUser *kc.UserRepresentation) bool {
 // AccreditationsProcessor interface
 type AccreditationsProcessor interface {
 	HasActiveAccreditations() bool
-	AddAccreditation(creationDate time.Time, name string, validity string)
-	RevokeAll() bool
-	RevokeTypes(accreditationsTypes []string) bool
+	AddAccreditation(creationDate time.Time, name string, validity string) AccreditationRepresentation
+	RevokeAll(notif func(AccreditationRepresentation)) bool
+	RevokeTypes(accreditationsTypes []string, notif func(AccreditationRepresentation)) bool
```
basbeu, line 68: why returning a bool when it is never used ?

internal/keycloakb/accreditationsmodule.go
```diff
@@ -117,19 +127,21 @@ func (ap *accredsProcessor) AddAccreditation(creationDate time.Time, name string
...
-		if ap.RevokeType(k) {
+		if ap.RevokeType(k, notif) {
 			res = true
 		}
 	}
 	return res
 }
 
-func (ap *accredsProcessor) RevokeType(accreditationsType string) bool {
+func (ap *accredsProcessor) RevokeType(accreditationsType string, notif func(AccreditationRepresentation)) bool {
```
basbeu, line 144: should not be a private method ?

basbeu: or exposed in the interface ?

pkg/validation/component.go
```diff
@@ -60,27 +60,33 @@ type Component interface {
...
-	originEvent         string
+	keycloakClient       KeycloakClient
+	tokenProvider        TokenProvider
+	archiveDBModule      ArchiveDBModule
+	eventReporterModule  EventsReporterModule
+	accredsService       AccreditationsServiceClient
+	configDBModule       ConfigurationDBModule
+	logger               log.Logger
+	originEvent          string
+	unknownAgentRealm    string
```
basbeu, line 71: Why not to use constants instead of field of a struct ?

fperot74: I just copied what was done for unknownAgentUserID and unknownAgentUsername... You ask yourself why you did that for those 2 fields! ;-)

basbeu (approved):

PR #498, opened 2024-04-12
[CLOUDTRUST-5396] Username not correctly initialized

fperot74:


pkg/management/component.go
```diff
@@ -432,8 +432,8 @@ func (c *component) genericCreateUser(ctx context.Context, accessToken string, c
 	}
 
 	var username = ""
-	if user.Username != nil {
-		username = *user.Username
+	if userRep.Username != nil {
```
basbeu, line 435: Was this causing a bug ? 

I see there is other places in this function in which we are using user instead of userRep.

basbeu (approved):

pkg/management/component.go
fperot74: Not a bug but we still had events without username when there is no reason not to have it

PR #499, opened 2024-05-01
[CLOUDTRUST-5408] Support multiple back urls

basbeu:


api/account/api.go
```diff
@@ -100,7 +100,7 @@ type Configuration struct {
 	AutoIdentificationVoucherEnabled      *bool           `json:"auto_identification_voucher_enabled"`
 	AutoIdentificationAccountingEnabled   *bool           `json:"auto_identification_accounting_enabled"`
 	AutoIdentificationPrepaymentRequired  *bool           `json:"auto_identification_prepayment_required"`
-	AllowedBackURL                        *string         `json:"allowed_back_url"`
+	AllowedBackURLs                       []string        `json:"allowed_back_urls"`
```
fperot74, line 103: Do we want to support only new version? Selfservice will be ready in the same time?

api/management/api.go
```diff
@@ -212,7 +212,7 @@ type RealmCustomConfiguration struct {
 	OnboardingClientID                  *string   `json:"onboarding_client_id"`
 	SelfRegisterGroupNames              *[]string `json:"self_register_group_names"`
 	BarcodeType                         *string   `json:"barcode_type"`
-	AllowedBackURL                      *string   `json:"allowed_back_url"`
+	AllowedBackURLs                     []string  `json:"allowed_back_urls"`
```
fperot74, line 215: Do we want to support only new version? Backoffice will be ready in the same time?

fperot74 (approved):

api/account/api.go
basbeu: Selfservice and Backoffice have been updated

api/management/api.go
basbeu: Selfservice and Backoffice have been updated

PR #502, opened 2024-07-10
[CLOUDTRUST-5450] Performance enhancement when updating authorizations

fperot74:


pkg/management/component.go
```diff
@@ -2069,6 +2085,21 @@ func (c *component) checkAllowedTargetRealmsAndGroupNames(ctx context.Context, r
 			}
 		}
 
+		/* REMOVE THIS : START */
```
basbeu, line 2088: normal, that it is still there ?

internal/keycloakb/configdbauthtransaction.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (at *authorizationTransaction) executeQuery(query string, args ...any) error {
+	_, err := at.tx.Exec(query, args...)
+	return err
+}
+
+func (at *authorizationTransaction) CreateAuthorizations(ctx context.Context, authz []configuration.Authorization) error {
+	// 50 rows per actual DB request
+	return at.createAuthorizationsMax(ctx, authz, 50)
```
basbeu, line 46: Should the max row be a parameter that we manually want to tweak (in config.yml file)? In order to not change the code if we want to change that.

internal/keycloakb/configdbmodule.go
```diff
@@ -89,6 +101,11 @@ func NewConfigurationDBModule(db sqltypes.CloudtrustDB, logger log.Logger, actio
 	}
 }
 
+func (c *configurationDBModule) execNoResult(query string, args ...any) error {
```
basbeu, line 104: I do not really see the benefit of this function. For me, this kind of wrapping, do not bring something to the code

pkg/management/authorizationdiff.go
```diff
@@ -0,0 +1,98 @@
...
+type AuthorizationSet interface {
+	apply(other AuthorizationSet, addFunction AlterFunction, removeFunction AlterFunction)
+}
+
+type diffStatus struct {
+	value configuration.Authorization
+}
+
+type authorizationSet struct {
+	authz map[string]any
```
basbeu, line 24: as we are dealing with authorization, do we really need a generic map ?

internal/keycloakb/configdbmodule.go
fperot74: Not a big benefit... just easier to be able to directly return an error rather than affecting it to a variable while ignoring the first returned value then returning the error

pkg/management/authorizationdiff.go
fperot74: It allows to write recursive generic code rather than code specific to each level of mapping. It was first written in a non-generic way and required almost 5 times more lines of code, just for copied/pasted code with a very small change at each time. For unit tests, it would have been necessary to write 5 times more tests.

pkg/management/component.go
fperot74: No

internal/keycloakb/configdbauthtransaction.go
```diff
@@ -0,0 +1,83 @@
...
+		args = append(args, nullableString(auth.RealmID), nullableString(auth.GroupName), nullableString(auth.Action),
+			nullableString(auth.TargetRealmID), nullableString(auth.TargetGroupName))
+	}
+
+	return at.executeQuery(query, args...)
+}
+
+func (at *authorizationTransaction) RemoveAuthorizations(ctx context.Context, authz []configuration.Authorization) error {
+	for _, auth := range authz {
+		if err := deleteAuthorization(at.executeQuery, *auth.RealmID, *auth.GroupName, *auth.TargetRealmID, auth.TargetGroupName, *auth.Action); err != nil {
```
harture, line 78: It seems strange to me to call an internal method of configdbmodule from here.

internal/keycloakb/configdbauthtransaction.go
```diff
@@ -0,0 +1,83 @@
...
+	// 50 rows per actual DB request
+	return at.createAuthorizationsMax(ctx, authz, 50)
+}
+
+func (at *authorizationTransaction) createAuthorizationsMax(ctx context.Context, authz []configuration.Authorization, maxRow int) error {
+	if len(authz) == 0 {
+		return nil
+	}
+	if len(authz) > maxRow {
+		if err := at.CreateAuthorizations(ctx, authz[0:maxRow]); err != nil {
```
harture, line 54: To make it more readable, the recursive method should call itself directly

internal/keycloakb/configdbauthtransaction.go
```diff
@@ -0,0 +1,83 @@
...
+			return err
+		}
+		return at.CreateAuthorizations(ctx, authz[maxRow:])
+	}
+	var valuesKeyword = "VALUES "
```
harture, line 59: I would suggest to extract this traetment logic in a submethod

internal/keycloakb/configdbauthtransaction.go
```diff
@@ -0,0 +1,83 @@
+package keycloakb
```
harture, line 1: Separation of duty between this file and configdbmodule is unclear to me. I would suggest to move all the authorizations queries is a dedicated file whcih would also contains the mechanism for transactions

harture (changes requested):

pkg/management/component.go
```diff
@@ -253,6 +254,7 @@ func NewComponent(keycloakClient KeycloakClient, kcURIProvider kc.KeycloakURIPro
 		authChecker:               authChecker,
 		tokenProvider:             tokenProvider,
 		authorizedTrustIDGroups:   authzedTrustIDGroups,
+		authorizationSetGenerator: newAuthorizationSet,
```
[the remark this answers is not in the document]
fperot74: It's just a trick to make unit tests be easier to write. This field is not updated in production code.

pkg/management/authorizationdiff.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func toNonNil(value *string) string {
+	if value == nil {
+		return nilString
+	}
+	return *value
+}
+
+func insertIntoMap(authz map[string]any, auth configuration.Authorization, values ...string) {
```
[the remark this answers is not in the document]
fperot74: I added comments. Hope it will easier to understand.

internal/keycloakb/configdbauthtransaction.go
fperot74: It is an internal static function... not linked to configdbmodule.
It is function deleteAuthorization(_ dbExecutable, ...), not function (db *configDBModule) deleteAuthorization(...)

internal/keycloakb/configdbauthtransaction.go
fperot74: ConfigDBModule is a module for queries without transactions... Requests are executed directly on a DB connection instance.
When creating a DB transaction, you want to execute requests on the transaction, not directly on the db connection.
Alternatives could be renaming configdbauthtransaction to configdbtransaction (it could execute other requests in a transaction, not only authorizations)... or have an instance of ConfigDBModule that wraps a first one (like context.Context), the wrapped one would use the transaction, the non-wrapped one would call directly the DB without transaction

harture: It seems to me there is some Transaction management in ConfigDBModule. Maybe we can see it more as batch queries ?

internal/keycloakb/configdbauthtransaction.go
harture: Fine, if we see it as an extension of ConfigDBModule

pkg/management/authorizationdiff.go
```diff
@@ -0,0 +1,136 @@
+package management
```
basbeu, line 1: There is a lot of generic names in this file : input, map, apply but we are dealing with authorization only. I think to have less generic name could make the code easier to understand

pkg/management/authorizationdiff.go
```diff
@@ -0,0 +1,136 @@
...
+	return authz
+}
+
+func newAuthorizationSet(input []configuration.Authorization) AuthorizationSet {
+	return &authorizationSet{
+		authz: toMap(input),
+	}
+}
+
+func (d *authorizationSet) Apply(otherIntf AuthorizationSet, addFunction AlterFunction, removeFunction AlterFunction) {
```
basbeu, line 99: why not call it Diff instead of vague apply ?

pkg/management/authorizationdiff.go
```diff
@@ -0,0 +1,136 @@
...
+	case map[string]any:
+		for _, subvalue := range v {
+			d.alterAuthorizations(subvalue, alterFunction)
+		}
+	case configuration.Authorization:
+		alterFunction(v)
+	}
+}
+
+func (d *authorizationSet) applyMap(thisMap map[string]any, otherMap map[string]any, addFunction AlterFunction, removeFunction AlterFunction) {
```
basbeu, line 115: why passing this map instead of doing d.authz ?

basbeu: I think the answer is at line 124 ;)

basbeu: but then I think that both alterAuthorizations and  applyMap could be simple function and not method of the struct ! right ?

pkg/management/authorizationdiff.go
basbeu: and instead of other, could we call other new ? because it is the new set of authorization. I think it could make easier to read

pkg/management/authorizationdiff.go
```diff
@@ -0,0 +1,136 @@
...
+		authz: toMap(input),
+	}
+}
+
+func (d *authorizationSet) Apply(otherIntf AuthorizationSet, addFunction AlterFunction, removeFunction AlterFunction) {
+	var other = otherIntf.(*authorizationSet)
+	d.applyMap(d.authz, other.authz, addFunction, removeFunction)
+}
+
+func (d *authorizationSet) alterAuthorizations(value any, alterFunction AlterFunction) {
```
harture, line 104: Do we really need to have two recursive methods (alterAuthorization and applyMap) for this diff operation ?
It gives me a strange feeling about the pattern.

pkg/management/component.go
```diff
@@ -225,6 +225,7 @@ type component struct {
 	authChecker               AuthorizationChecker
 	tokenProvider             toolbox.OidcTokenProvider
 	authorizedTrustIDGroups   map[string]bool
+	authorizationSetGenerator func([]configuration.Authorization) AuthorizationSet
```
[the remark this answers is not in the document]
fperot74: It was for testing purpose in the non-generic way of processing the map. It is not necessary anymore with the generic version. Removed

pkg/management/authorizationdiff.go
fperot74: Fixed

pkg/management/authorizationdiff.go
fperot74: Hmmm? They are not doing the same processing... One is searching for differences, the other one knows that a given part of the map only contains differences.

harture (approved):

PR #504, opened 2024-09-23
[CLOUDTRUST-5400] Move GLN validation to custom ct-gln validator

jonas-tr:


cmd/keycloakb/keycloak_bridge.go
```diff
@@ -532,6 +532,8 @@ func main() {
 	}
 	var glnVerifier = business.NewGlnVerifier(glnLookupProviders...)
 
+	profile.GlnVerifier = glnVerifier
```
fperot74, line 535: I'm not fan of static assigning variables from other packages.
You should pass glnVerifier to each "object" you are creating and which will need it

jonas-tr: After talking to Francis, we concluded that the current solution was fine.

basbeu (approved):

PR #505, opened 2024-10-08
[CLOUDTRUST-5515] Allow more inputs for advanced search

Xelowak:
Keycloak-bridge used to deny and return an invalid parameter error if we search for usernames, first/last names, or emails that do not satisfy a RegExp. This is not the behavior we intend as it may be useful to search for an address containing a substring.

### DIFF
--- a/pkg/management/http.go
+++ b/pkg/management/http.go
@@ -68,7 +68,7 @@ func MakeManagementHandler(e endpoint.Endpoint, logger log.Logger) *http_transpo
 
 // decodeEventsRequest gets the HTTP parameters and body content
 func decodeManagementRequest(ctx context.Context, req *http.Request) (interface{}, error) {
-	var pathParams = map[string]string{
+	pathParams := map[string]string{
 		prmRealm:        constants.RegExpRealmName,
 		prmUserID:       constants.RegExpID,
 		prmClientID:     constants.RegExpClientID,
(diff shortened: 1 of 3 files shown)


pkg/management/http.go
```diff
@@ -79,11 +79,11 @@ func decodeManagementRequest(ctx context.Context, req *http.Request) (interface{
 		prmAction:       constants.RegExpName,
 	}
 
-	var queryParams = map[string]string{
-		prmQryEmail:         constants.RegExpEmail,
+	queryParams := map[string]string{
+		prmQryEmail:         constants.RegExpSearch,
 		prmQryFirstName:     constants.RegExpSearch,
-		prmQryLastName:      constants.RegExpLastNameSearch,
```
fperot74, line 85: So RegExpLastNameSearch and RegExpUsername are still in use?

fperot74 (approved):

pkg/management/http.go
Xelowak: Apparently, RegExpLastNameSearch can be safely removed but RegExpUsername is used [here](https://github.com/cloudtrust/keycloak-bridge/blob/master/pkg/kyc/http.go).

harture (approved):

harture:
We have to determine why CI is failing before merging it

basbeu (approved):

basbeu:
TestHTTPManagementHandler/Invalid_input_parameter is failing, please update it !

basbeu (changes requested):
TestHTTPManagementHandler/Invalid_input_parameter is failing, please update it !

basbeu (approved):

PR #512, opened 2024-10-31
[CLOUDTRUST-5540] Onboarding status feature flag

Xelowak:


pkg/management/component.go
```diff
@@ -391,7 +391,17 @@ func (c *component) genericCreateUser(ctx context.Context, accessToken string, c
 	user api.UserRepresentation, generateUsername bool, generateNameID bool, termsOfUse bool, useOnboardingCheckForExistingUser bool) (string, error) {
 	var userRep = api.ConvertToKCUser(user)
 	userRep.SetAttributeString(constants.AttrbSource, source)
-	userRep.SetAttributeString(constants.AttrbOnboardingStatus, managementOnboardingStatus)
+
+	realm := ctx.Value(cs.CtContextRealm).(string)
```
harture, line 395: I think we should use the targetRealmName. Realm from context could be "master" when you create a user into a customer realm

pkg/register/component.go
```diff
@@ -323,7 +324,16 @@ func (c *component) createUser(ctx context.Context, accessToken string, realmNam
 	kcUser.Groups = &groupIDs
 
 	// Set onboarding status
-	kcUser.SetAttributeString(constants.AttrbOnboardingStatus, registerOnboardingStatus)
+	realm := ctx.Value(cs.CtContextRealm).(string)
```
harture, line 327: Use the realmName parameter.
Not even sure the realmName is available in the context ofr the register.

harture (changes requested):

pkg/management/component_test.go
```diff
@@ -515,6 +523,14 @@ func TestCreateUser(t *testing.T) {
 
 		var onboardingStatus = "user-created-by-api"
 
+		var ctx = context.WithValue(context.Background(), cs.CtContextAccessToken, accessToken)
+		ctx = context.WithValue(ctx, cs.CtContextRealm, realmName)
```
basbeu, line 527: I think with your changes after the review, you do not need to set the context anymore

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22
 
 require (
 	github.com/IBM/sarama v1.41.1
-	github.com/cloudtrust/common-service/v2 v2.7.0
+	github.com/cloudtrust/common-service/v2 v2.7.1
```
basbeu, line 7: Looks like last version is [v2.8.1](https://github.com/cloudtrust/common-service/releases/tag/v2.8.1)

pkg/management/component.go
```diff
@@ -391,7 +391,15 @@ func (c *component) genericCreateUser(ctx context.Context, accessToken string, c
 	user api.UserRepresentation, generateUsername bool, generateNameID bool, termsOfUse bool, useOnboardingCheckForExistingUser bool) (string, error) {
 	var userRep = api.ConvertToKCUser(user)
 	userRep.SetAttributeString(constants.AttrbSource, source)
-	userRep.SetAttributeString(constants.AttrbOnboardingStatus, managementOnboardingStatus)
+
+	realmAdminConfig, err := c.configDBModule.GetAdminConfiguration(ctx, targetRealmName)
+	if err != nil {
+		c.logger.Info(ctx, "err", err.Error())
```
harture, line 397: If there is an error, we should return.

harture: Warn seems to me more appropriate to log this error.
Maybe also add an explicit message, that we cannot retireve realmAdmin configuration  for example.

pkg/register/component.go
```diff
@@ -323,7 +323,14 @@ func (c *component) createUser(ctx context.Context, accessToken string, realmNam
 	kcUser.Groups = &groupIDs
 
 	// Set onboarding status
-	kcUser.SetAttributeString(constants.AttrbOnboardingStatus, registerOnboardingStatus)
+	_, realmAdminConfig, err := c.configDBModule.GetConfigurations(ctx, realmName)
+	if err != nil {
+		c.logger.Info(ctx, "err", err.Error())
```
harture, line 328: Same as for management about return and log message

harture (changes requested):
Sorry I missed those point in my previous review ...

go.mod
Xelowak: Yes but this release updates dependencies and we do not want to so it yet

basbeu (approved):

harture (approved):

PR #527, opened 2024-12-20
[CLOUDTRUST-5718] Allow empty strings as redirect URIs

Xelowak:


Xelowak:
We decided to resolve this issue on the frontend side

PR #530, opened 2025-01-06
[CLOUDTRUST-5722] Allow cancellation of email/phone change

Xelowak:


pkg/account/component.go
```diff
@@ -541,3 +543,51 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
...
+
+	// Get the "old" user representation from Keycloak
+	oldUserKc, err := c.keycloakAccountClient.GetAccount(accessToken, realm)
+	if err != nil {
+		c.logger.Warn(ctx, "msg", "Can't get account", "err", err.Error())
+		return err
+	}
+	keycloakb.ConvertLegacyAttribute(&oldUserKc)
+
+	oldUserKc.SetAttributeString(constants.AttrbEmailToValidate, "")
```
fperot74, line 559: Even if user profiles might force an empty string, oldUserKc.RemoveAttribute(...) would make more sens.

pkg/account/component.go
```diff
@@ -541,3 +543,51 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
...
+
+	// Get the "old" user representation from Keycloak
+	oldUserKc, err := c.keycloakAccountClient.GetAccount(accessToken, realm)
+	if err != nil {
+		c.logger.Warn(ctx, "msg", "Can't get account", "err", err.Error())
+		return err
+	}
+	keycloakb.ConvertLegacyAttribute(&oldUserKc)
+
+	oldUserKc.SetAttributeString(constants.AttrbPhoneNumberToValidate, "")
```
fperot74, line 583: Even if user profiles might force an empty string, oldUserKc.RemoveAttribute(...) would make more sens.

fperot74, line 583: What if phoneNumberToValidate is already missing/blank? Should we update such an account?

pkg/account/component.go
fperot74, line 559: What if emailToValidate is already missing/blank? Should we update such an account?

pkg/account/component.go
```diff
@@ -541,3 +543,51 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
 
 	return err
 }
+
+func (c *component) CancelEmailChange(ctx context.Context) error {
```
fperot74, line 547: You also could have something like:

func (c *component) CancelEmailChange(ctx context.Context) error {
    return c.removeAttribute(ctx, constants.AttrbEmailToValidate)
}
func (c *component) CancelPhoneNumberChange(ctx context.Context) error {
    return c.removeAttribute(ctx, constants.AttrbPhoneNumberToValidate)
}
func (c *component) removeAttribute(...) error {
    ...
}

pkg/account/component.go
```diff
@@ -541,3 +543,51 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
...
+
+	oldUserKc.SetAttributeString(constants.AttrbEmailToValidate, "")
+
+	// Update keycloak account
+	err = c.keycloakAccountClient.UpdateAccount(accessToken, realm, oldUserKc)
+	if err != nil {
+		c.logger.Warn(ctx, "msg", "Can't update account", "err", err.Error())
+		return err
+	}
+
```
fperot74, line 567: Missing event...
See event from UpdateAccount: c.eventReporterModule.ReportEvent(ctx, events.NewEventOnUserFromContext(ctx, c.logger, c.originEvent, "UPDATE_ACCOUNT", realm, userID, username, nil))

basbeu (changes requested):
Please make the Francis suggestions !

pkg/account/component.go
Xelowak: Thanks, I didn't realize it existed!

pkg/account/component.go
Xelowak: Probably not if it's missing, you're right.
But if it's an empty string, we could remove it from the user.

I'll make the necessary changes

pkg/account/component.go
```diff
@@ -541,3 +543,44 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
...
+func (c *component) CancelEmailChange(ctx context.Context) error {
+	return c.removeStringAttributeFromUser(ctx, constants.AttrbEmailToValidate)
+}
+
+func (c *component) CancelPhoneNumberChange(ctx context.Context) error {
+	return c.removeStringAttributeFromUser(ctx, constants.AttrbPhoneNumberToValidate)
+}
+
+func (c *component) removeStringAttributeFromUser(ctx context.Context, attr kc.AttributeKey) error {
```
fperot74, line 555: No need to specify StringAttribute... We don't care if it contains a string or not.

pkg/account/component.go
```diff
@@ -541,3 +543,44 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
...
+
+	// Get the user representation from Keycloak
+	userKc, err := c.keycloakAccountClient.GetAccount(accessToken, realm)
+	if err != nil {
+		c.logger.Warn(ctx, "msg", "Can't get account", "err", err.Error())
+		return err
+	}
+	keycloakb.ConvertLegacyAttribute(&userKc)
+
+	if userKc.GetAttributeString(attr) == nil {
```
fperot74, line 567: Did you tested with user profile activated on the realm? I wonder if "missing" attribute is nil or empty with user profile.

pkg/account/component.go
```diff
@@ -541,3 +543,44 @@ func (c *component) executeActions(ctx context.Context, actions []string) error
...
+	userKc.RemoveAttribute(attr)
+
+	// Update keycloak account
+	err = c.keycloakAccountClient.UpdateAccount(accessToken, realm, userKc)
+	if err != nil {
+		c.logger.Warn(ctx, "msg", "Can't update account", "err", err.Error())
+		return err
+	}
+
+	userID := ctx.Value(cs.CtContextUserID).(string)
```
fperot74, line 581: Are we sure to have them in the context? (perhaps populated when checking JWT)
I would just use values from userKc...

fperot74 (approved):

pkg/account/component.go
Xelowak: Are you sure about this? I'm using userKc.GetAttributeString below to check if the attribute is not present, so I guess I need to specify it

pkg/account/component.go
Xelowak: From my understanding, the function `GetString` from `model_toolbox.go` (in keycloak-client) that is called by `GetAttributeString` already transforms an empty string into a nil, so I don't think I need to treat the empty string case here.

Here is the function as a reference:
```
// GetString gets the first value of a given attribute
func (a Attributes) GetString(key AttributeKey) *string {
	var attrbs = a[key]
	if len(attrbs) == 1 && attrbs[0] == "" {
		// User profile sometimes sets a default empty value
		return nil
	}
	if len(attrbs) > 0 {
		return &attrbs[0]
	}
	return nil
}
```

pkg/account/component.go
basbeu: Yes we are sure. It is the way to retrieve userID in the account api. We do even like this in the update endpoint.

We go through a middleware MakeHTTPOIDCTokenValidationMW that validates the token and put some values in the context.

basbeu (approved):

pkg/account/component.go
fperot74: All attributes are strings (more precisely, string arrays)

pkg/account/component.go
fperot74: Ok... Taking them from userKc is just a direct access to the values

pkg/account/component.go
Xelowak: Ok, then I'll change this

PR #536, opened 2025-01-23
[CLOUDTRUST-5782] Add identification URI endpoint

Xelowak:
We need to merge and release https://github.com/cloudtrust/common-service/pull/160 first, then I will update dependencies (go.mod, go.sum, vendor/modules.txt) there.


pkg/configuration/component.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *component) GetIdentificationURI(ctx context.Context, realm string, contextKey string) (string, error) {
+	ctxOverride, ok := c.contextKeyMgr.GetOverride(realm, contextKey)
+	if !ok {
+		c.logger.Info(ctx, "msg", "Invalid (context-key, realm) pair", "context-key", contextKey, "realm", realm)
+		return "", errorhandler.CreateBadRequestError(errorhandler.MsgErrInvalidParam + ".context-key")
+	}
+
+	if ctxOverride.IdentificationURI == nil {
+		c.logger.Info(ctx, "msg", "Empty identification URI", "context-key", contextKey, "realm", realm)
+		return "", errorhandler.CreateBadRequestError(errorhandler.MsgErrInvalidParam + ".context-key")
```
Xelowak, line 42: I'm not sure about those bad request errors, so don't hesitate to suggest other ideas.

pkg/configuration/http_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+		assert.Nil(t, err)
+		assert.Equal(t, http.StatusOK, res.StatusCode)
+
+		body, err := io.ReadAll(res.Body)
+		assert.Nil(t, err)
+
+		// Remove quotes from the server response
+		result := strings.Trim(string(body), "\"")
+		assert.Equal(t, identificationURI, result)
```
Xelowak, line 48: If you have another solution, please tell me.

pkg/configuration/component.go
```diff
@@ -0,0 +1,46 @@
...
+		logger:        logger,
+	}
+}
+
+// GetIdentificationURI returns the identification URI for a given context key and realm
+func (c *component) GetIdentificationURI(ctx context.Context, realm string, contextKey string) (string, error) {
+	ctxOverride, ok := c.contextKeyMgr.GetOverride(realm, contextKey)
+	if !ok {
+		c.logger.Info(ctx, "msg", "Invalid (context-key, realm) pair", "context-key", contextKey, "realm", realm)
+		return "", errorhandler.CreateBadRequestError(errorhandler.MsgErrInvalidParam + ".context-key")
```
fperot74, line 37: The pair realm/contextKey is incorrect... So why not? You also can use ".realm-context-key" or ".realm-and-context-key"...

pkg/configuration/component.go
fperot74: It's not a context-key related error... This could be a 5xx error as there is a problem on our side because realm is not correctly configured.
400 Bad request => invalid-param.no-uri?
412 Precondition Failed? Precondition would be realm should be configured first? => precondition.realm-configuration
417 Expectation failed? (almost the same as 412) => expectation.realm-configuration

fperot74: Another way of solving this question, at startup, if identification URI is not provided, make the service start fails... (are there cases where we don't want to configure it?)

pkg/configuration/http.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	prmContextKey = "context-key"
+	prmRealmName  = "realm"
+)
+
+// MakeConfigurationHandler makes an HTTP handler for the configuration endpoint
+func MakeConfigurationHandler(e endpoint.Endpoint, logger log.Logger) *http_transport.Server {
+	return http_transport.NewServer(e,
+		decodeConfigurationRequest,
```
fperot74, line 22: You can directly inline decodeConfigurationRequest here.
Ex: https://github.com/cloudtrust/keycloak-bridge/blob/master/pkg/kyc/http.go#L28

pkg/configuration/http_test.go
fperot74: You might be able to use json.Unmarshal(...)... but for this simple test, this way of processing the string does not shock me.
But if you follow my previous comment, you would not have to process the result the same way.
You also could have tests where you check what happens if context-key is not a UUID

pkg/configuration/http_test.go
```diff
@@ -0,0 +1,50 @@
...
+	mockConfigurationComponent := mock.NewComponent(mockCtrl)
+
+	const (
+		realm             = "trustid"
+		contextKey        = "19251660-f869-11ec-b939-0242ac120002"
+		identificationURI = "http://identification-uri"
+	)
+
+	r := mux.NewRouter()
+	r.Handle("/configuration/realms/{realm}/identification", MakeConfigurationHandler(keycloakb.ToGoKitEndpoint(MakeGetIdentificationURIEndpoint(mockConfigurationComponent)), log.NewNopLogger()))
```
fperot74, line 30: You don't want to test all this package code at once. In unit test, you can write a simple handler that concatenate the realm and the context key or write them as a JSON map. This way, you can check that your HTTPConfigurationHandler correctly identify the realm and the context-key and you can check its behavior when the provided context-key does not match the UUID regex

fperot74 (changes requested):

pkg/configuration/component.go
Xelowak: I'll look into it, thanks for the suggestions

go.mod
```diff
@@ -79,3 +79,5 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudtrust/common-service/v2 => ../../common-service
```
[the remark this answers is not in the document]
Xelowak: As I said in the description of my PR, it is to remind me that we must first merge and release my PR on common-service, then I can update the dependencies here.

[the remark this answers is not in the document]
basbeu: I do not understand why you need an update of common-service

pkg/configuration/component.go
Xelowak: I decided to go with 404 to emphasize that we do not have an identification URI to provide for this (realm, context-key) pair

basbeu: I would not make this service fails at startup for bad context-key configuration. I think it is kind of not flexible, we want maybe configure at the end for all customer. But maybe not at the same time.

For me it is an error part of 500 errors family

fperot74 (approved):

harture (approved):

PR #557, opened 2025-03-31
[CLOUDTRUST-5585] Return 404 when updating user with invalid UID on KYC

jmettler27:


Xelowak:
Can you add a test case for this?

pkg/kyc/component.go
```diff
@@ -408,6 +408,12 @@ func (c *component) validateUser(ctx context.Context, accessToken string, confRe
 	kcUser, err = c.keycloakClient.GetUser(accessToken, targetRealm, userID)
 	if err != nil {
 		c.logger.Warn(ctx, "msg", "Can't get user/accreditations", "err", err.Error())
+		switch e := err.(type) {
```
basbeu, line 411: Please add a unit test covering this scenario. 

And is there more elegant to use errors.As method ? more information https://go.dev/blog/go1.13-errors

basbeu (changes requested):

pkg/kyc/component.go
jmettler27: Other methods use the same pattern with the switch case, like [here](https://github.com/cloudtrust/keycloak-bridge/blob/master/pkg/kyc/component.go#L226) and [here](https://github.com/cloudtrust/keycloak-bridge/blob/master/pkg/support/component.go#L41), and `errors.As` are not used.

fperot74: errors.As suggested by @basbeu seems a more elegant way of testing the error type especially here where we only are interrested by one case.

jmettler27: Okay thanks! I changed it.

pkg/kyc/component.go
```diff
@@ -408,6 +409,10 @@ func (c *component) validateUser(ctx context.Context, accessToken string, confRe
 	kcUser, err = c.keycloakClient.GetUser(accessToken, targetRealm, userID)
 	if err != nil {
 		c.logger.Warn(ctx, "msg", "Can't get user/accreditations", "err", err.Error())
+		var httpErr kc.HTTPError
+		if errors.As(err, &httpErr) && httpErr.HTTPStatus == 404 {
+			return errorhandler.CreateNotFoundError("user")
```
fperot74, line 414: I would use a c.logger.Info(...) here to tell the user does not exist and use c.logger.Warn(...) only in other cases.
Originally, we decided to log with Error(...) for errors that need an urgent action and log with Warn(...) when action has to be taken with no emergency.

fperot74 (approved):

pkg/kyc/component.go
jmettler27: Okay thanks, I commited those changes. Did I understood correctly?

PR #560, opened 2025-04-03
[CLOUDTRUST-6058] Include attributes when getting realm role

Xelowak:


fperot74 (approved):
Don't forget to update the swagger file before merging

pkg/management/component.go
```diff
@@ -1572,6 +1572,7 @@ func (c *component) GetRole(ctx context.Context, realmName string, roleID string
 	roleRep.ClientRole = roleKc.ClientRole
 	roleRep.ContainerID = roleKc.ContainerID
 	roleRep.Description = roleKc.Description
+	roleRep.Attributes = roleKc.Attributes
```
basbeu, line 1575: As discussed, please filter the businessRoleFlag

basbeu (changes requested):

pkg/management/component.go
```diff
@@ -1573,6 +1573,12 @@ func (c *component) GetRole(ctx context.Context, realmName string, roleID string
 	roleRep.ContainerID = roleKc.ContainerID
 	roleRep.Description = roleKc.Description
 
+	// Filter out the business role flag
+	attributes := roleKc.Attributes
+	delete(*attributes, "BUSINESS_ROLE_FLAG")
```
fperot74, line 1578: Are we sure there is no case where roleKc.Attributes is nil?

fperot74: Not sure why you are using the temporary variable attributes. It is the same pointer to a map, not a new map. I guess you can directly write:
roleRep.Attributes = roleKc.Attributes
delete(*roleRep.Attributes, "BUSINESS_ROLE_FLAG")
(perhaps with a test if roleRep.Attributes==nil)

basbeu (approved):

PR #562, opened 2025-04-04
[CLOUDTRUST-6058] Update attributes when updating realm role

Xelowak:


pkg/management/component.go
```diff
@@ -1635,6 +1635,11 @@ func (c *component) UpdateRole(ctx context.Context, realmName string, roleID str
 	kcRole.ContainerID = role.ContainerID
 	kcRole.Description = role.Description
 
+	// We update the attributes but keep the business role flag
+	businessRoleFlagValue := (*kcRole.Attributes)[businessRoleFlag]
```
fperot74, line 1639: What if we don't have this attribute?

Xelowak: It means you're trying to do something really weird, which is updating a non-business role by calling the bridge endpoint. Normally you shouldn't even be aware of non-business roles, as it's impossible to retrieve them using the bridge.

fperot74: I just checked in production...
select kr.ID, kr.NAME ROLE_NAME, kr.REALM_ID, ra.ROLE_ID, ra.NAME, ra.VALUE
from KEYCLOAK_ROLE kr
left outer join ROLE_ATTRIBUTE ra ON ra.ROLE_ID=kr.ID

2080 roles, only 5 "BUSINESS_ROLE_FLAG"... You have to check if this attribute exists or not

fperot74 (changes requested):

pkg/management/component.go
Xelowak: Yes, I'm aware a lot of roles do not have this attribute, but if it's the case I'm pretty sure they should not be updated using the bridge.
Anyway, I'll add this to make sure we don't fall into stupid errors

fperot74 (approved):

PR #576, opened 2025-04-30
[CLOUDTRUST-6173] Return roleID in location when creating a role

Xelowak:


pkg/management/component.go
```diff
@@ -1601,13 +1601,26 @@ func (c *component) CreateRole(ctx context.Context, realmName string, role api.R
 		return "", err
 	}
 
-	//retrieve the role ID
-	reg := regexp.MustCompile(`[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}`)
-	roleID := string(reg.Find([]byte(locationURL)))
+	roles, err := c.keycloakClient.GetRoles(accessToken, realmName)
+	if err != nil {
+		c.logger.Warn(ctx, "err", err.Error())
```
fperot74, line 1606: Warn(ctx, "msg", "Can't get roles from Keycloak", "err", err.Error())

pkg/management/component.go
```diff
@@ -1601,13 +1601,26 @@ func (c *component) CreateRole(ctx context.Context, realmName string, role api.R
...
+		return "", err
+	}
+
+	// Search for role ID
+	var roleID string
```
fperot74, line 1611: It would be better to ensure you found the correct role (even if there is no reason not to find it)...
You could have something like:
if roleID, err = c.getRoleIDFromName(roles, *role.Name); err!=nil { ... }

fperot74 (approved):

PR #605, opened 2025-07-15
[CLOUDTRUST-6642] SSE API: retrieve and delete IDP links

jmettler27:


go.mod
```diff
@@ -65,13 +65,17 @@ require (
...
-	golang.org/x/text v0.25.0 // indirect
+	golang.org/x/text v0.26.0 // indirect
 	google.golang.org/protobuf v1.36.5 // indirect
 	gopkg.in/go-jose/go-jose.v2 v2.6.3 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudtrust/keycloak-client/v2 => ../keycloak-client
+
+replace github.com/cloudtrust/common-service/v2 => ../common-service
```
jmettler27, line 81: For local testing. Will update the PR once keycloak-client and common-service are released.

basbeu: I understand it is very convenient for local testing. However it does not run in the CI. Do you know you can take module based on a branch ? this way will work in the CI (if the branch is pushed on github for sure)

for example : 
go get github.com/cloudtrust/common-service/v2@MyBranch

jmettler27: Ah I didn't know, I fixed it.

fperot74: Better than a branch, you can use a commit (when pushing, you have a message with xxyyzzold...aabbccnew). Its easier with commit ids because if you push on a branch, go mod vendor does not see all changes

pkg/account/component_test.go
```diff
@@ -1219,3 +1223,90 @@ func TestRemoveAttributeFromUser(t *testing.T) {
...
+	var providerAlias = "idp"
+	var anyError = errors.New("any error")
+
+	var ctx = context.WithValue(context.Background(), cs.CtContextAccessToken, accessToken)
+	ctx = context.WithValue(ctx, cs.CtContextRealm, realmName)
+	ctx = context.WithValue(ctx, cs.CtContextUserID, userID)
+	ctx = context.WithValue(ctx, cs.CtContextUsername, username)
+
+	t.Run("Delete linked account with succces", func(t *testing.T) {
+		mocks.keycloakAccountClient.EXPECT().DeleteLinkedAccount(accessToken, realmName, providerAlias).Return(nil)
```
fperot74, line 1298: Here, it's easier to understand... Context has been initialized earlier, we quickly see that the difference is the return anyError in the fail case when success case return nil followed by another call to a mock

fperot74 (approved):

PR #620, opened 2025-08-04
[CLOUDTRUST-6417] Add attributes to IdentityProviderRepresentation

jmettler27:
Add `hideOnLogin`


api/idp/api.go
```diff
@@ -8,56 +8,62 @@ import (
...
+	Enabled                     *bool                   `json:"enabled,omitempty"`
+	FirstBrokerLoginFlowAlias   *string                 `json:"firstBrokerLoginFlowAlias,omitempty"`
+	HideOnLogin                 *bool                   `json:"hideOnLogin,omitempty"`
+	InternalID                  *string                 `json:"internalId,omitempty"`
+	LinkOnly                    *bool                   `json:"linkOnly,omitempty"`
+	PostBrokerLoginFlowAlias    *string                 `json:"postBrokerLoginFlowAlias,omitempty"`
+	ProviderID                  *string                 `json:"providerId,omitempty"`
+	StoreToken                  *bool                   `json:"storeToken,omitempty"`
+	TrustEmail                  *bool                   `json:"trustEmail,omitempty"`
+	UpdateProfileFirstLoginMode *string                 `json:"updateProfileFirstLoginMode,omitempty"`
```
fperot74, line 25: Are you considering values on/off rather than true/false?

jmettler27: Yes, the two expected values for this are "on"/"missing"/"off", and in our [template ](https://projectportal.elca.ch/jira/browse/CLOUDTRUST-6568) we set it to "on" for all our IDPs. But actually, this attribute is marked as deprecated in the [docs ](https://www.keycloak.org/docs-api/latest/javadocs/org/keycloak/representations/idm/IdentityProviderRepresentation.html#updateProfileFirstLoginMode)... So with @harture we decided to remove this attribute for now

PR #623, opened 2025-08-06
[CLOUDTRUST-6576] Add endpoints to get/update Theme configuration and Theme translation

jonas-tr:


vendor/modules.txt
```diff
@@ -8,7 +8,7 @@ github.com/IBM/sarama
 # github.com/VividCortex/gohistogram v1.0.0
 ## explicit
 github.com/VividCortex/gohistogram
-# github.com/cloudtrust/common-service/v2 v2.11.2
+# github.com/cloudtrust/common-service/v2 v2.11.3-0.20250806103933-1eba84ab5caf
```
basbeu, line 11: do not forget to merge and release first common service

pkg/management/component.go
```diff
@@ -209,6 +211,10 @@ type Component interface {
 	UnlinkShadowUser(ctx context.Context, realmName string, userID string, provider string) error
 
 	GetIdentityProviders(ctx context.Context, realmName string) ([]api.IdentityProviderRepresentation, error)
+
+	GetThemeConfiguration(ctx context.Context, realmName string) (api.ThemeConfiguration, error)
+	UpdateThemeConfiguration(ctx context.Context, realmName string, themeConf api.UpdatableThemeConfiguration) error
+	GetThemeTranslation(ctx context.Context, realmName string, language string) (interface{}, error)
```
basbeu, line 217: replace interface{} by any . But can't we have a stricter data format ? 

I think we need a stricter format, it is also related to the lack of validation pointed in api.

pkg/management/component.go
```diff
@@ -209,6 +211,10 @@ type Component interface {
 	UnlinkShadowUser(ctx context.Context, realmName string, userID string, provider string) error
 
 	GetIdentityProviders(ctx context.Context, realmName string) ([]api.IdentityProviderRepresentation, error)
+
+	GetThemeConfiguration(ctx context.Context, realmName string) (api.ThemeConfiguration, error)
```
basbeu, line 215: General question, did we ask for endpoint in management API ? or did we discuss about that ?

pkg/management/component.go
```diff
@@ -2558,3 +2564,119 @@ func (c *component) GetIdentityProviders(ctx context.Context, realmName string)
...
+	err := c.configDBModule.UpdateThemeConfiguration(ctx, config)
+	if err != nil {
+		c.logger.Warn(ctx, "err", err.Error())
+		return err
+	}
+
+	return nil
+}
+
+func (c *component) GetThemeTranslation(ctx context.Context, realmName string, language string) (interface{}, error) {
```
basbeu, line 2656: Why do we need an endpoint specialized for retrieving translations ?

basbeu (changes requested):

pkg/management/component.go
jonas-tr: I'm not really sure what stricter format would work with the translation files

pkg/management/component.go
jonas-tr: It's because the library used for the translation loads the translation only if needed. So when we change language it calls the endpoint and get the translation file to use

api/management/api.go
```diff
@@ -343,6 +355,22 @@ type IdentityProviderRepresentation struct {
...
+// ThemeConfiguration struct
+type ThemeConfiguration struct {
+	Settings *configuration.ThemeConfigurationSettings `json:"settings,omitempty"`
+	Logo     *string                                   `json:"logo,omitempty"`
+	Favicon  *string                                   `json:"favicon,omitempty"`
+}
+
+// UpdatableThemeConfiguration struct
+type UpdatableThemeConfiguration struct {
+	ThemeName    *string                                   `json:"theme_name,omitempty"`
```
fperot74, line 367: We could have a unique struct ThemConfiguration with the ThemeName, Settings, Translations, Logo and FavIcon... (which are "nilable")
UpdatableXXX struct where introduced in some of our interfaces because we add to face an issue when some users of the interface where transmitting null values in the JSON to tell they wanted no change in an update statement when we were using to null value to clear a field in database. We don't have this problem here

api/management/api.go
```diff
@@ -343,6 +355,22 @@ type IdentityProviderRepresentation struct {
...
+	Settings     *configuration.ThemeConfigurationSettings `json:"settings,omitempty"`
+	Translations *map[string]any                           `json:"translations,omitempty"`
+	Logo         *[]byte                                   `json:"logo,omitempty"`
```
fperot74, line 370: slices can be nil... Don't use a pointer here, it adds complexity with indirections.
We used to use pointer but we are trying to remove them when not required

api/management/api.go
```diff
@@ -343,6 +355,22 @@ type IdentityProviderRepresentation struct {
...
+	Favicon      *[]byte                                   `json:"favicon,omitempty"`
```
fperot74, line 371: slices can be nil... Don't use a pointer here, it adds complexity with indirections.
We used to use pointer but we are trying to remove them when not required

pkg/management/component.go
```diff
@@ -2559,3 +2564,78 @@ func (c *component) GetIdentityProviders(ctx context.Context, realmName string)
...
+		c.logger.Warn(ctx, "err", err.Error())
+		return "", err
+	}
+
+	if backofficeConfig.BoTheme == nil {
+		c.logger.Warn(ctx, "msg", "No theme configured for realm")
+		return "", errorhandler.CreateNotFoundError("theme")
+	}
+
+	translationStr, err := c.configDBModule.GetThemeTranslation(ctx, *backofficeConfig.BoTheme, language)
```
fperot74, line 2628: Should we force language to be lowercase or uppercase to avoid inconsistencies in database?

api/management/api.go
```diff
@@ -1228,3 +1256,82 @@ func ConvertToAPIUserChecks(checks []accreditationsclient.CheckRepresentation) [
...
+		ValidateParameterMaxSize(constants.Logo, themeConf.Logo, maxLogoImageSize, true).
+		ValidateParameterImageMimeType(constants.Logo, themeConf.Logo, allowedLogoImageTypes, true).
+		ValidateParameterMaxSize(constants.Favicon, themeConf.Favicon, maxFaviconImageSize, true).
+		ValidateParameterImageMimeType(constants.Favicon, themeConf.Favicon, allowedFaviconImageTypes, true).
+		ValidateParameterMaxSize(constants.Translations, &translationsJSON, maxTranslationsSize, true).
+		ValidateParameterOnlyStrings(constants.Translations, themeConf.Translations, true).
+		Status()
+}
+
+func getImageMimeType(imageData []byte) (string, error) {
```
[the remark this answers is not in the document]
basbeu: I agree with Francis, we should have this code to only one place

PR #632, opened 2025-08-21
[CLOUDTRUST-6827] Add per-API read-only user profile annotation

Xelowak:


internal/profile/validation.go
```diff
@@ -70,7 +80,8 @@ func Validate(ctx context.Context, upc UserProfile, realm string, input Contains
 func ValidateUser(profile kc.UserProfileRepresentation, input ContainsFields, apiName string, checkMandatory bool) error {
 	for _, attrb := range profile.Attributes {
 		if !attrb.AnnotationMatches(apiName, func(value string) bool {
-			return strings.EqualFold(value, "true") || strings.EqualFold(value, "required")
+			// Account is the only API supporting read-only
+			return strings.EqualFold(value, "true") || strings.EqualFold(value, "required") || (apiName == "account" && strings.EqualFold(value, "read-only"))
```
fperot74, line 84: Not sure if the bridge should do this check... The bridge is a pass-through, it is the responsibility of the one doing the configuration and the one coding the self-service to do it correctly.

api/common/api-profile.go
```diff
@@ -77,19 +78,24 @@ func AttributeToAPI(attrb kc.ProfileAttrbRepresentation, apiName string) *Profil
 		// Component should not be aware of its existence
 		return nil
 	}
-	// By default, attributes are not shown for a given frontend
+
+	// By default, attributes are not shown for a given frontend. Account is the only API supporting read-only
 	if apiName != "" && (!attrb.AnnotationMatches(apiName, func(value string) bool {
-		return strings.EqualFold(value, "true") || strings.EqualFold(value, "required")
+		return strings.EqualFold(value, "true") || strings.EqualFold(value, "required") || (apiName == "account" && strings.EqualFold(value, "read-only"))
```
fperot74, line 84: Not sure if the bridge should do this check... The bridge is a pass-through, it is the responsibility of the one doing the configuration and the one coding the self-service to do it correctly.

basbeu (approved):

PR #637, opened 2025-08-25
[CLOUDTRUST-6842] Allow to set user password in account API

Xelowak:


pkg/account/component.go
```diff
@@ -132,17 +133,47 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
 	var userID = ctx.Value(cs.CtContextUserID).(string)
 	var username = ctx.Value(cs.CtContextUsername).(string)
 
-	if currentPassword == newPassword || newPassword != confirmPassword {
-		return errorhandler.Error{
-			Status:  http.StatusBadRequest,
-			Message: keycloakb.ComponentName + "." + "invalidValues",
+	if currentPassword == "" {
+		// Current password is empty, meaning the user wants to set their password. Then, we need to check they do not have a password credential
```
fperot74, line 137: the user ... his/her password

pkg/account/component.go
```diff
@@ -132,17 +133,47 @@ func (c *component) UpdatePassword(ctx context.Context, currentPassword, newPass
...
 		}
-	}
 
-	_, err := c.keycloakAccountClient.UpdatePassword(accessToken, realm, currentPassword, newPassword, confirmPassword)
-	if err != nil {
-		c.logger.Warn(ctx, "msg", "Can't update password", "err", err.Error())
-		return err
+		for _, cred := range credentialsKc {
+			if *cred.Type == "password" {
+				c.logger.Warn(ctx, "msg", "Can't set password", "err", "user has a password credential")
```
fperot74, line 146: I think "Can't set initial password as user already owns one" would be clearer ("err" is not a mandatory tag)

pkg/account/component_test.go
```diff
@@ -67,37 +67,79 @@ func TestUpdatePassword(t *testing.T) {
...
+	t.Run("Set password: GetCredentials fails", func(t *testing.T) {
+		mocks.keycloakAccountClient.EXPECT().GetCredentials(accessToken, realm).Return([]kc.CredentialRepresentation{}, anError)
+
+		err := component.UpdatePassword(ctx, "", newPasswd, newPasswd)
+		assert.NotNil(t, err)
+	})
+
+	t.Run("Set password: user has a current password", func(t *testing.T) {
+		cred := kc.CredentialRepresentation{
+			Type: cs.ToStringPtr("password"),
```
fperot74, line 116: ToStringPtr is a complex processing used for incoming data according to a given user profile. We don't really need this so complex processing... but why not.

fperot74 (approved):

pkg/account/component_test.go
Xelowak: You're right, I changed it

PR #648, opened 2025-09-16
[CLOUDTRUST-6903] Add HRD management for IDPs

jmettler27:
Changes:
- Modified CRUD of an IDP to account for its (optional) HRD settings - managed via KC's Components API. In the end we removed the components API of the bridge and instead moved the components logic to the bridge. This is done via a ComponentTool.
- IDP representation: Changed type of the map attribute (Config); Increase length of Alias regex


internal/constants/regexp.go
```diff
@@ -63,11 +64,19 @@ const (
 	RegExpCorporateLastName  = regExpLen128
 
 	// Identity provider
-	RegExpAlias                     = RegExpRealmName
+	RegExpAlias                     = `^EXTIDP-[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}$`
```
fperot74, line 67: EXTIDP- should be specific to external idp service. The bridge should not be aware of this specific format.
Currently, identity providers are used only for AGW but it could be any identity provider so there is no reason for the alias to only start with EXTIDP-

pkg/account/authorization_test.go
```diff
@@ -119,7 +119,7 @@ func TestDeny(t *testing.T) {
 
 	var falseBool = false
 	var credentialID = "786-5684-6464"
-	var providerAlias = "idp"
+	var providerAlias = "EXTIDP-12345678-abcd-efgh-ijkl-012345678901"
```
fperot74, line 122: Cancel this change

pkg/account/authorization_test.go
```diff
@@ -177,7 +177,7 @@ func TestAllowed(t *testing.T) {
 
 	var trueBool = true
 	var credentialID = "786-5684-6464"
-	var providerAlias = "idp"
+	var providerAlias = "EXTIDP-12345678-abcd-efgh-ijkl-012345678901"
```
fperot74, line 180: Cancel this change

pkg/account/authorization_test.go
```diff
@@ -239,7 +239,7 @@ func TestError(t *testing.T) {
 	var realmName = "master"
 
 	var credentialID = "786-5684-6464"
-	var providerAlias = "idp"
+	var providerAlias = "EXTIDP-12345678-abcd-efgh-ijkl-012345678901"
```
fperot74, line 242: Cancel this change

pkg/account/component_test.go
```diff
@@ -1327,7 +1327,7 @@ func TestGetLinkedAccounts(t *testing.T) {
 	var currentRealm = "test-community"
 	var anyError = errors.New("any error")
 
-	var providerAlias = "idp"
+	var providerAlias = "EXTIDP-12345678-abcd-efgh-ijkl-012345678901"
```
fperot74, line 1330: Cancel this change

fperot74 (changes requested):

pkg/components/http_test.go
```diff
@@ -0,0 +1,59 @@
...
+	compsHandler2 := MakeComponentsHandler(keycloakb.ToGoKitEndpoint(MakeGetComponentsEndpoint(mockComponent)), mockLogger)
+
+	r := mux.NewRouter()
+	r.Path("/components/realms/{realm}/components").Methods("POST").Handler(compsHandler)
+	r.Path("/components/realms/{realm}/components").Methods("GET").Handler(compsHandler2)
+
+	ts := httptest.NewServer(r)
+	defer ts.Close()
+
+	{
```
fperot74, line 39: You can name your sub tests using t.Run("sub name", func(t *testing.T) { ... })

fperot74 (approved):

Xelowak (approved):

pkg/idp/component_test.go
```diff
@@ -16,35 +16,72 @@ import (
...
+func createTestKcIdp() kc.IdentityProviderRepresentation {
 	return kc.IdentityProviderRepresentation{
-		AddReadTokenRoleOnCreate:  ptrBool(false),
-		Alias:                     ptr(idpAlias),
-		AuthenticateByDefault:     ptrBool(false),
-		Config:                    &map[string]interface{}{},
+		AddReadTokenRoleOnCreate: ptrBool(false),
+		Alias:                    ptr(idpAlias),
+		AuthenticateByDefault:    ptrBool(false),
+		Config: map[string]any{
```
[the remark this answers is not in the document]
fperot74: Do we really need all these values for test purpose? It seems you use some real values but we usually don't need so "rich" data for tests.

api/idp/api.go
```diff
@@ -70,7 +76,6 @@ func (idp IdentityProviderRepresentation) Validate() error {
 		ValidateParameterRegExp("alias", idp.Alias, constants.RegExpAlias, true).
 		ValidateParameterRegExp("displayName", idp.DisplayName, constants.RegExpDisplayName, true).
 		ValidateParameterRegExp("firstBrokerLoginFlowAlias", idp.FirstBrokerLoginFlowAlias, constants.RegExpFirstBrokerLoginFlowAlias, true).
-		ValidateParameterRegExp("internalId", idp.InternalID, constants.RegExpID, true).
```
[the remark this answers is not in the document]
basbeu: additional question why do we do no validate every field of the api ?

configs/keycloak_bridge.yml
```diff
@@ -187,6 +187,21 @@ onboarding-realm-overrides:
   realmName-hp: realmName
   realmName-institution: realmName
 
+context-keys:
```
basbeu, line 190: why do you add context-keys ?

basbeu (approved):

api/idp/api.go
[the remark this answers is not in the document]
jmettler27: I added validation for the config map and the HRD settings. For the internalId, I put it as optional.

PR #678, opened 2025-11-14
[CLOUDTRUST-7424] Use pointer values for HRD settings

jmettler27:


api/idp/api.go
```diff
@@ -28,8 +28,8 @@ type IdentityProviderRepresentation struct {
 }
 
 type HrdSettingModel struct {
-	IPRangesList string `json:"ipRangesList"`
-	Priority     int    `json:"priority"`
+	IPRangesList *string `json:"ipRangesList"`
```
fperot74: I think we should have field domainsList to avoid losing its value if it is already set.

api/idp/api.go
```diff
@@ -28,8 +28,8 @@ type IdentityProviderRepresentation struct {
...
+	Priority     *int    `json:"priority"`
```
fperot74: IPRangesList and DomainsList should be optional but why Priority is not mandatory?

PR #680, opened 2025-11-18
[CLOUDTRUST-6953] Adapt AGW ID to hex value of 128 bits

jmettler27:


fperot74 (approved):

basbeu (commented):
Is it normal we only have changes in test ? was it to correct a unit test ?

PR #748, opened 2026-04-07
[CLOUDTRUST-8029] Add an endpoint to know if the user can get identified

Xelowak:


basbeu (changes requested):

api/management/api_test.go
```diff
@@ -701,6 +701,7 @@ func TestConvertRealmAdminConfiguration(t *testing.T) {
 			ConsentRequiredSocial:                          boolPtr(true),
 			ConsentRequiredCorporate:                       boolPtr(false),
 			ConsentRequiredCorporateAuxiliary:              boolPtr(false),
+			AccreditationRenewalWindowDays:                 intPtr(30),
```
fperot74: Didn't we say that we should start using the new way of getting pointers from constants, coming with the latest Golang version?
If so, it would be better to remove boolPtr/intPtr/... functions, at least in this package and use the new way. (perhaps AI could help applying it to the whole project?)

pkg/account/component.go
```diff
@@ -671,3 +675,78 @@ func (c *component) DeleteLinkedAccount(ctx context.Context, providerAlias strin
 
 	return nil
 }
+
+func (c *component) GetCanIdentify(ctx context.Context, contextKey *string) (bool, error) {
+	var accessToken = ctx.Value(cs.CtContextAccessToken).(string)
+	currentRealm := ctx.Value(cs.CtContextRealm).(string)
+
+	accreditationType := "DEP"
```
fperot74: Do we really want an hardcoded value for the accreditation type?

pkg/account/component.go
```diff
@@ -671,3 +675,78 @@ func (c *component) DeleteLinkedAccount(ctx context.Context, providerAlias strin
...
+	renewalWindowDays := c.defaultRenewalWindowDays
+	if adminConfig.AccreditationRenewalWindowDays != nil {
+		renewalWindowDays = *adminConfig.AccreditationRenewalWindowDays
+	}
+	deadline := time.Now().AddDate(0, 0, renewalWindowDays)
+
+	hasAccreditationOfType := false
+	latestExpiration := time.Time{}
+	for _, accred := range accreditations {
+		if accred.Type != nil && *accred.Type == accreditationType && !*accred.Revoked {
```
fperot74: !*accred.Revoked => (accred.Revoked==null || !*accred.Revoked)

Also, you could write a `function (acc Accreditation) CheckIsActive() error` in api : easier to test the whole check, have a smaller code here and easier tests for this part.

pkg/account/endpoint_test.go
```diff
@@ -209,6 +209,20 @@ func TestMakeDeleteLinkedAccountEndpoint(t *testing.T) {
 	assert.Nil(t, err)
 }
 
+func TestMakeGetCanIdentifyEndpoint(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	mockAccountComponent := mock.NewComponent(mockCtrl)
+	mockAccountComponent.EXPECT().GetCanIdentify(gomock.Any(), gomock.Any()).Return(true, nil) // TODO: verify the context key is passed correctly
```
fperot74: TODO?
You can write 2 tests, one with map[string]string{}, the other with map[string]string{ prmQueryContextKey: "context-key" }

fperot74 (approved):

pkg/account/authorization.go
```diff
@@ -244,3 +245,8 @@ func (c *authorizationComponentMW) checkAPISelfIDPLinksManagementEnabled(ctx con
 
 	return c.checkFlagEnabled(ctx, config.APISelfIDPLinksManagementEnabled, action, "Forbidden: linked accounts management disabled")
 }
+
+func (c *authorizationComponentMW) GetCanIdentify(ctx context.Context, contextKey *string) (bool, error) {
```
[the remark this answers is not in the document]
Xelowak: Yes, I think that's the point

configs/keycloak_bridge.yml
```diff
@@ -217,3 +217,5 @@ kafka:
     initial-offset: newest
 
 log-events-rate: 1
+
+default-renewal-window-days: 30
```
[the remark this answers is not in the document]
Xelowak: You're right, I removed the harcoded part in the api

api/management/api.go
```diff
@@ -899,6 +911,7 @@ func ConvertRealmAdminConfigurationFromDBStruct(conf configuration.RealmAdminCon
 		ConsentRequiredSocial:                          defaultBool(conf.ConsentRequiredSocial, false),
 		ConsentRequiredCorporate:                       defaultBool(conf.ConsentRequiredCorporate, false),
 		ConsentRequiredCorporateAuxiliary:              defaultBool(conf.ConsentRequiredCorporateAuxiliary, false),
+		AccreditationRenewalWindowDays:                 defaultInt(conf.AccreditationRenewalWindowDays, 30),
```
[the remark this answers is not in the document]
Xelowak: I changed it to 60 days

api/management/api_test.go
Xelowak: We already have a different Jira for running `go fix` on all repositories so I prefer not to modify the old code in this PR. 
I changed my line to use `new`

pkg/account/component.go
Xelowak: I tried it for a while but could not find an easier way of writing this function using a utility function. The thing is I'm not really searching for this information but rather for the latest expiration time of a specific type of accreditation, which is harder to modularize.

Also, I updated the condition based on your remark of `Revoked` being nullable.

pkg/account/component.go
Xelowak: I added it inside configuration

pkg/account/endpoint_test.go
Xelowak: Thanks for pointing this out!
I indeed forgot to update my test before pushing

vendor/modules.txt
```diff
@@ -5,7 +5,7 @@ filippo.io/edwards25519/field
 # github.com/IBM/sarama v1.47.0
 ## explicit; go 1.25.0
 github.com/IBM/sarama
-# github.com/cloudtrust/common-service/v2 v2.18.1
+# github.com/cloudtrust/common-service/v2 v2.18.2-0.20260402083019-bd82fd649fd2
```
basbeu: Do not forget to first merge and release common service

configs/keycloak_bridge.yml
```diff
@@ -217,3 +217,6 @@ kafka:
     initial-offset: newest
 
 log-events-rate: 1
+
+default-accreditation-requested: "DEP"
+default-renewal-window-days: 60
```
basbeu: For deployment, I would rather at least set 62, because two monthes could be 62 days. Why not setting 65 days to have margin and avoid weird edge case.

basbeu (approved):

configs/keycloak_bridge.yml
Xelowak: Okay, I change it to 65

harture (approved):

PR #784, opened 2026-07-02
[CLOUDTRUST-8787] Add context key in EMAIL_ONBOARDING_SENT event

Xelowak:


fperot74 (approved):
Would it be too long to have one test with and one without contextKey ?

Xelowak:
> Would it be too long to have one test with and one without contextKey ?

Yes, I'm not sure the amount of effort I would put into this would be worth it. I checked if I could easily test this, but we only check for `ReportEvent(gomock.Any(), gomock.Any())` calls, which means I would have to add a lot of logic just for this.

pkg/management/component.go
```diff
@@ -1300,7 +1300,11 @@ func (c *component) genericSendOnboardingEmail(ctx context.Context, accessToken
 	}
 
 	// store the API call into the DB
-	c.auditEventsReporterModule.ReportEvent(ctx, events.NewEventOnUserFromContext(ctx, c.logger, c.originEvent, "EMAIL_ONBOARDING_SENT", realmName, userID, *kcUser.Username, nil))
+	details := map[string]string{}
+	if contextKey != nil {
+		details["contextKey"] = *contextKey
```
basbeu: in details of events we are not consistent with naming. Sometimes camel case sometimes snake case. However I think we could have other events in which we want to set the contextKey in the future. Could make a constant somewhere to try to not be inconsistent on this one ? 

I've checked we do not have the context key in other events yet.

basbeu (approved):

PR #820, opened 2026-09-15
[CLOUDTRUST-8889] Fix sponsors attribute value

fperot74:


pkg/validation/sponsors.go
```diff
@@ -0,0 +1,44 @@
+package validation
```
basbeu: We usually do not set this kind of file/module in pkg. 

It looks like to me to an internal module. Why not having a internal module for that ? 

Knowing that in the future other api than validation will need to have access to this attribute too.

basbeu (approved):

